PR #10, opened 2026-04-08
fix: resolve three CI test failures on main

pghege:
## Summary
- **time**: use `total_seconds()` instead of `int64` cast for pandas 2.x compatibility (1000x scaling error)
- **ingest**: lowercase `cell_id` when creating nested metadata directories
- **tests**: generate legacy parquet fixture inline instead of referencing missing file

## Test plan
- [x] All 37 unit tests pass locally
- [x] CI passes on this branch

PR #14, opened 2026-04-11
Refactor: split __init__.py and enforce plugin interface contracts

### LINKED ISSUE #11: Refactor monolithic __init__.py and add plugin contract validation
## Summary

`src/bdf/__init__.py` is currently ~2,095 lines and serves as the entire public API surface. As the package grows, this will become a maintenance bottleneck. Additionally, the plugin system lacks compile-time or import-time validation that format plugins implement the required interface.

### Split `__init__.py` into submodules

The file currently contains:
- Dataset helpers (`load_registry`, `get_entry`, `search`, `sparql`)
- Cleaning utilities (`clean`, `CleanReport`)
- Visualization wrappers (`plot`, `explore`)
- Ingestion logic (`ingest`)
- Source resolution (local files, URLs, remote registries)
- Template generation

These are logically distinct concerns and could live in their own modules with re-exports from `__init__.py` to preserve the public API (`from bdf import read, validate, clean`, etc.).

### Add plugin contract validation

`CyclerPlugin` subclasses are auto-registered via metaclass, but there is no check that a plugin implements the required methods (`sniff`, `parse`). A plugin missing a method will fail at runtime with an `AttributeError` rather than a clear error at import time.

Options:
- Use `abc.abstractmethod` on the base class
- Add a metaclass `__init_subclass__` check that raises `TypeError` with a descriptive message

## Motivation

Both changes reduce the cost of adding new cycler formats and make the codebase easier to navigate for new contributors.

pghege:
## Summary

Closes #11

- **Plugin contract enforcement**: Added `abc.abstractmethod` decorators to `CyclerPlugin.sniff()` and `CyclerPlugin.parse()` in `base.py`. Subclasses that fail to implement these required methods now raise `TypeError` at import time instead of silently inheriting no-op stubs. All existing plugins already implement both methods, so no plugin changes were needed.

- **Split monolithic `__init__.py`**: Extracted two new internal modules:
  - `src/bdf/_source.py` — source resolution logic (`_is_url`, `_resolve_source`, `_looks_like_bdf_artifact`, `_candidate_plugins`)
  - `src/bdf/_ingest.py` — the ~1200-line `ingest()` function and all its nested helpers (GitHub download, DOI enrichment, metadata generation, state tracking, etc.)

- **Public API unchanged**: All re-exports in `__init__.py` are preserved. `from bdf import read, validate, clean, ingest` etc. continue to work exactly as before.

## Test plan

- [x] `pytest tests/unit -q` — 35 passed, 2 pre-existing failures (unrelated to this change, tracked in #10)
- [ ] CI green on this branch
- [ ] Verify `from bdf import read, validate, clean, ingest, detect, plugins` works in a fresh interpreter


pghege:
Superseded by the 0.2.0 API freeze work: Graham has the read/scan split and load removal, and the module split is tracked on the org roadmap for after the freeze. This diff predates the spec-driven refactor (it edits the removed data_sources/ modules), so closing rather than rebasing.

PR #18, opened 2026-05-07
Add Biologic mpr, bugfix OCV mpt

### LINKED ISSUE #15: Read Biologic raw mpr using `yadg`
[yadg](https://github.com/dgbowl/yadg) already has an mpr -> xarray/netcdf parser which could be used for an mpr -> bdf

I can't assign myself, but happy to give this a go

g-kimbell:
Adds a biologic mpr parser using [yadg](https://github.com/dgbowl/yadg), closes #15 

I'm not sure how you'd prefer data to be stored, I see a lot is being grabbed from zenodo instead of being added to the repo. There's ~30 MB of data here zipped to ~8 MB.

This PR also fixes an issue reading OCV mpt files, which do not include a current column, that I found while debugging the mprs.


jsimonclark:
Thanks for working on this issue, @g-kimbell ! Reviewing your changes, I see only some minor things that would need to be resolved. 

But I think the main question is if we are allowed to use yadg because of their license conditions. I see they use [GPL-3](https://github.com/dgbowl/yadg?tab=GPL-3.0-1-ov-file#readme) which is a copyleft license that requires derived works to also be licensed GPL-3. That conflicts with the more permissive [Apache-2](https://github.com/battery-data-alliance/battery-data-format#Apache-2.0-1-ov-file) license of BDF. 

So, unfortunately I doubt we can use it as a dependency due to the distribution clause. @pghege , what is your opinion?

g-kimbell:
I don't think there's a big licence risk using yadg as an optional dependency, as we're not distributing the yadg source code, but I'll double check with Peter.

There could be an issue putting the test data in this repo. Peter said he had no objections, but the actual data is provided by several people over the last years (https://github.com/dgbowl/yadg/issues/256). I could write a small script to download the data from yadg repo instead when tests are run, then we never distribute them.

g-kimbell:
I've removed the yadg data from the tests/data folder

Instead, the test data is downloaded with `requests` and cached in `~/.cache/bdf/biologic`

For licences, it should not be an issue if we keep yadg as an optional dependency, and never distribute any yadg source code - see https://github.com/dgbowl/yadg/issues/256

One remaining issue is Python versions - yadg>=6 only supports Python>=3.10. We can add this as an optional dependency with 
```
[project.optional-dependencies]
yadg = ["yadg>=6.2; python_version >= '3.10'"]
```
and skip yadg tests in the CI for Python 3.9.

(We have the same issue with `fastnda`)

g-kimbell:
Tests fail because we don't have optional dependencies installed in the CI. We cannot install yadg (or fastnda) on Python 3.9.

I think this failure happens on main and isn't related to this PR:
`FileNotFoundError: [Errno 2] No such file or directory: 'data/empa__ccid000001.bdf.parquet'`

g-kimbell:
This was for 0.1.x, but has gone stale with all the recent changes. Starting a new PR for new read methods.

PR #20, opened 2026-05-13
Fix/step quantity semantics

jsimonclark:
## Summary

Resolves semantic issues raised during review of the BDF specification. Corrects broken column mappings across all plugins, introduces two new columns, and defines derivation formulas for all six capacity/energy quantities.

## New columns

- `step_id` (`Step ID / 1`) — schedule step identifier as reported by the instrument (Arbin   `Step_Index`, Neware `Step_ID`, Digatron `Step`, BioLogic `Ns`). May be non-contiguous and  repeats across cycles.  
- `step_type` (`Step Type / 1`) — step operational mode string (e.g. `CC_CHG`, `CC_DCH`, `REST`). Vocabulary not yet  standardised; values preserved as reported.

## Major changes

**Column mapping corrections (all plugins)**
All plugins previously mapping a schedule step ID to `Step Index / 1` or `Step Count / 1` now map to `Step ID / 1`: Digatron, Neware NDA, Neware CSV, LANDT CSV/TXT, Novonix. BioLogic `Ns` added to `Step ID / 1`. `step_count` synonyms cleared — it is derived-only, no longer populated from vendor files. Neware CSV and LANDT cycle columns corrected to `Cycle Count / 1`.

**Capacity/energy definitions**
- `cumulative_capacity_ah` / `_energy_wh` , now correctly throughput (`charging + discharging`), always non-decreasing. Previously emitted as net by the Digatron plugin.  
- `net_capacity_ah` / `_energy_wh` , running integral from test start (`charging − discharging`), can be negative. Equivalent to BioLogic Q−Q₀. 
- `charging_capacity_ah`, `discharging_capacity_ah` and energy equivalents, clarified as test-level accumulators that never reset between steps or cycles.

**Digatron plugin**
`fixup()` rewritten to derive all six secondary columns from the two primary accumulators, replacing instrument-internal `AhAccu`, `AhBal`, `WhAccu`, `WhBal` values.

**Cycle Count starting value**
Clarified: any instrument-defined starting value (0, 1, or user-configured) is valid and must be preserved. Cycle 0 typically denotes pre-cycling steps.


g-kimbell:
Hi Simon, nice changes and clarifications :)

One small issue is  `Step Type / 1` implies to me it's a dimensionless numerical quantity rather than a category.

In case you are collecting them, for Neware all the step types I know of are here https://github.com/EmpaEconversion/fastnda/blob/main/fastnda/dicts.py#L9

jsimonclark:
@g-kimbell thanks, you are right! I removed the unit from step_type and step_id. I'm still working to integrate some comments from @pghege and will update the PR.

PR #28, opened 2026-06-03
Add legacy data

g-kimbell:
Adds a (cut down) legacy bdf file for tests

Fixes `test_legacy_labels_normalized_from_ontology` failing


g-kimbell:
I think the CI failure is happening on main

Maybe we should just merge #10 which resolves this legacy data anyway

PR #38, opened 2026-06-10
Feat/ontology driven docs

### LINKED ISSUE #36: Conflicting `step_capacity_ah` definitions
The readme has a conflicting definition of `step_capacity_ah` with the ontology. The readme says:

> Net capacity change over a given step

i.e. it is signed

The ontology has said:

> Electric charge transferred during the current test step

Which is ambiguous, I would've expected signed, then later

> Running accumulation of charge throughput within the current test step, in ampere hour: the time integral of the absolute current from the step start

i.e. absolute, unsigned

fastnda 1.1.0 uses it to mean net capacity which I read from the readme. We should fix the readme and maybe change the ontology description to say it is ambiguous (and hence deprecated).

jsimonclark:
# Summary
Makes the BDF ontology snapshot the single source of truth for all term documentation and the required-column set, eliminating version drift.

Resolves #36 The conflicting `step_capacity_ah` definitions were this drift: the hand-maintained README described it as a signed "net capacity change" while the ontology defines the unsigned running accumulation (renamed to `step_cumulative_capacity_ah` in ontology 1.1.0, with `step_capacity_ah` deprecated). The README tables are now generated from the ontology, so the deprecated term and its wrong description are gone and this class of conflict can no longer occur , CI fails if the tables drift.

- **Quantity model** now carries `obligation`, `definition`, `description`, `latex_symbol`, `latex_formula`, and `derived_from` from the snapshot. Requiredness derives from the ontology's `:obligation` annotation (hardcoded set retained only as a pre-1.1.0 fallback); behavioral tests pin the required/recommended sets so an ontology release that reclassifies a column fails visibly instead of silently changing `validate_df()`.
- **README term tables** (Required/Recommended/Optional) are generated by `scripts/generate_docs.py` between `BEGIN/END GENERATED` markers, stamped with the source ontology version. First regeneration fixes accumulated drift (deprecated `step_capacity_ah` row, wrong descriptions, malformed `power_watt` link). CI fails if the tables drift (`--check`).
- **Release-pinned snapshot**: `get_snapshot(version=...)` / `bdf-update-snapshot 1.1.0` fetch a tagged ontology release and verify the fetched `owl:versionInfo` matches. `bdf --version` reports both package and snapshot versions.
- **Daily sync** (`sync-ontology.yml`): compares the latest ontology release to the bundled snapshot; no-op when equal, otherwise updates, regenerates, and opens a PR whose body is a term-level summary (added/removed/deprecated/obligation changes) — not a raw TTL diff — with an explicit warning when the required set changes.
- `CONTRIBUTING.md` documents the contract: term metadata changes go to the ontology repo; generated regions are never hand-edited.


jsimonclark:
@tomjholland I'm flagging this before merge since it overlaps with your work on `src/bdf/spec.py`. 

The goal is that the ontology is used everywhere as the source of truth to avoid version drift (i.e. no/minimal hardcoding of annotations).

The relevant change for you is commit 8abd220: Quantity gains ontology-sourced fields (obligation, definition, description, latex_symbol, latex_formula, derived_from), and required is now derived from the ontology's obligation annotation instead of the hardcoded set (pinned by `tests/unit/test_spec_ontology_fields.p`y). The README tables are also generated from ColumnOntology now, so those fields are load-bearing for scripts/generate_docs.py and a CI check.

Could you take a quick look at just that commit and say whether it conflicts with the direction of your `spec `refactor?

src/bdf/spec.py
```diff
@@ -581,34 +612,62 @@ def from_graph(cls, g: Any) -> "ColumnOntology":
...
+
+        # The ontology's :obligation annotation is the source of truth for
+        # requiredness. Pre-1.1.0 ontologies carry no obligations; only then
+        # does the static _REQUIRED_DEFAULT fallback (set per-subject above)
+        # remain in effect.
+        if any(q.obligation for q in quantities.values()):
+            quantities = {
+                name: q.model_copy(update={"required": q.obligation == "required"})
+                for name, q in quantities.items()
+            }
```
tomjholland: This can probably be removed if we get rid of `Quantity.required` as a field. If we want we can add a property to replace it that just checks the value of the `obligation` field, which will eliminate the chance of these ever being out-of-sync. This means all current call sites expecting a boolean returned by `.required` will be unaffected. 

The `obligation` field can be filled within `Quantity.from_graph_subject()` with a check for whether the ontology provides an `obligation` (or on ontology version number). If `obligation` is provided then that value is filled, otherwise it queries the hardcoded fallback set in `_REQUIRED_DEFAULT`

tomjholland (commented):
Looks good to me! Only one minor comment to avoid having two fields containing duplicate information that may be in conflict.

PR #41, opened 2026-06-16
Fix unitless quantities and temperature units

tomjholland:
This PR resolves a bug where unitless quantities like `step_id` and `step_type` weren't loaded into the `ColumnOntology` instance as they had no `/ {unit}`. These updates allow the `unit` for a quantity to be `None` and load the unit directly from the ontology instead of parsing the preferred label. The change also applies a fix for unit conversion when the symbol is given as `C`, which is interpreted as `degC` if the destination unit is a temperature unit.


jsimonclark:
@tomjholland CI was failing on two things: 

- the generated README table was out of sync with the new step_id/step_type quantities, and 
- _pint_understands let a non-UndefinedUnitError from pint escape (crashed import bdf.spec on Windows/pint≥0.25). 

I pushed 3 small commits: 

- README regen, 
- broadened the except to (PintError, AssertionError), and 
- made the ℃ test platform-independent.

Please check and confirm this is ok with you.

tomjholland:
> @tomjholland CI was failing on two things: 
> 
> 
> 
> - the generated README table was out of sync with the new step_id/step_type quantities, and 
> 
> - _pint_understands let a non-UndefinedUnitError from pint escape (crashed import bdf.spec on Windows/pint≥0.25). 
> 
> 
> 
> I pushed 3 small commits: 
> 
> 
> 
> - README regen, 
> 
> - broadened the except to (PintError, AssertionError), and 
> 
> - made the ℃ test platform-independent.
> 
> 
> 
> Please check and confirm this is ok with you. 

Yep looks good to me. Odd that pint produces bare assertion errors but this handles it well. All good to merge from my point of view.

PR #43, opened 2026-06-18
Spec driven reading and normalization

### LINKED ISSUE #23: Eliminate fixup()
Debugging normalisation is made trickier by the duplication of normalisation logic across `normalize()` and `fixup()`. #21 completely separates reading logic from normalisation so should resolve this.

### LINKED ISSUE #24: Add Arbin and Maccor parsers

tomjholland:
This PR introduces new modules to provide a unified object-oriented framework for normalising and reading battery cycler data into BDF format. This is intended to replace the existing backend of the `bdf.normalize()`, `bdf.detect()` and `bdf.read()` public-facing methods while allowing more advanced user interaction with the backend directly for processing new or custom formats. This PR introduces a more complete integration test suite that checks real cycler file cases on all of the new methods, to confirm the accuracy of the detection pipeline and correctness of bdf synonyms.

This PR only introduces new modules, with the existing methods remaining in place. A follow up PR will be made to replace user-facing methods and internal call-sites with the new methods, and remove deprecated code.

# New modules

## `bdf.table_normalizers` - commit 24fde303bdd6ca123d1ee60bc70e812d09d8f9a2
- A single point for all normalisation work, requiring no additional `fixup()`-like methods (closes #23)
- `TableNormalizer` class contains `Syn` and `DateTimeSyn` fields for each of the non-deprecated quantities in the ontology (confirmed with a test), which define cycler column synonyms
- Units are handled with a `{unit}` wildcard in `Syn` definitions for unit recognition and conversion (unit conversion executed within the `bdf.spec` module)
- `assumed` attribute of the `Syn` class is picked up by integration tests (described later) to mark synonyms that we don't have evidence for in our test dataset
- `TableNormaliser` singletons are defined for each of the major cycler types
- Module-level `bdf.table_normalizers.normalize()` method detects the correct normalizer from the dataframe columns, and calls its `TableNormalizer.normalize()` method 
- `normalize()` methods have `extra_columns` argument to allow non-bdf columns through the normalisation process and `validate` argument that checks the result against the ontology with the `spec` module's `ColumnOntology.validate()` method

## `bdf.table_parsers` - commit 172ba16ade9fd30808586494ed7716ce539eb4cc
- Contains all of the logic for reading (without normalising) data from raw files
- `TableParser` classes lightly wrap `polars` `read_*`/`scan_*` methods. As pydantic models they can be loaded from config files, instantiated in python or used in the `plugins` module (described later)
- Text-delimited file sniffing entirely structural; algorithm searches for the longest run of identically structured data e.g. (string, numerical, numerical, numerical, string) repeated n times to detect the header row and separator. This avoids the dependency on normalizer information that was previously used, where the header search was done against column synonyms.
- Daughter `TableParser`s made for delimited text, excel (using by default `fastexcel` which is polar's preferred engine, and is order-of-magnitude faster than `openpyexcel`), parquet, Matlab .mat (with scipy) and neware nda (using only `FastNDA`)
- Detection of table parsers works on file-extension and magic bytes for binary files
- `TableParser.read()` method first reads the raw data per the provided settings in the pydantic model instance, then passes to its `self.normalizer.normalize()` for normalisation (if not disabled)

## `bdf.metadata_parsers` - commit d3e9fb93a16d21ff6a10bf29cc8c1719998d90e9
- Structured similarly to `table_parsers`, with a `MetadataParser` class containing a `MetadataSchema` field
- `MetadataSchema` defines the extraction for metadata fields (analagous to the `TableNormalizer`). For its json variant, it defines simple mapping from json to bdf metadata fields. For text file headers it defines regex patterns for pulling metadata
- `start_time` added as the first "defined" metadata field that can be extracted in this way, with the aim (not implemented) to combine with `test_time_second` to auto-generate `unix_time_second` column where possible
- `TxtPreambleParser` includes `magic` field, responsible for identifying cycler files from strings in their header like `"basytec battery test system"`

## `bdf.plugins` - commit dc090e94c01bf5adc03d7da66889fcec60c3d516
- `Plugin` combines a `TableParser` and `MetadataParser`. Also a pydantic class so can be instantiated from config files and defined in python
- `bdf.plugins.PLUGINS` singleton is a dictionary of plugin string names to `Plugin` instances. Include fuzzy matching for error messages that suggest close matches
- `bdf.plugins.detect()` uses three stage pipeline for automatic plugin detection:
  - `detect_from_ext_or_magic_bytes` uses `TableParser` to check if the file extension or magic bytes at start match the file
  - `detect_from_metadata` uses `MetadataParser` magic strings (currently only for text files)
  - `detect_from_columns` uses `TableNormalizer` column synonym scoring
  - Includes plugins for Arbin and Maccor, closes #24 

## `bdf.io` - commit 07eecc9023b588568b6a7b2de9d7fa84c742f5b5
- Only modified file, provides home for `bdf.io.read()` (intended to replace `bdf.read()` in a later PR)
- `bdf.io.read()` now returns a `(df, metadata_dict)` tuple (will be breaking when `bdf.read()` is replaced)
- First calls `bdf.plugins.detect()` followed by `resolved_plugin.table_parser.read()` and `resolved_plugin.metadata_parser.parse()` to join the whole reading and normalisation chain

# Docs
- Examples included for the new modules and their classes (commit 5823b22a5342ec71f85c109387d12138594b2fd6)

# Integration tests
- `tests/integration/test_cases.py` defines expectations for test cases
- `tests/integration/test_table_parsers.py`, `tests/integration/test_metadata_parsers.py` and `tests/integration/test_plugins.py` use the cases to parameterise tests for their plugin detection, reading and file sniffing methods
- `tests/integration/test_table_normalizers.py` (commit 8d18ad251714fa70e2d1644e04c5ec870833c630) includes a parameterised test that ensures every synonym marked as `assumed=True` is present in the sample data and actively converted to a bdf column. Synonyms marked as `assumed=False` return `XFAIL` (expected fail) so don't block the CI, but feedback which synonyms are missing from the test data. If a synonym is incorrectly labelled either way the test case will fail
- `tests/integration/test_column_validity.py` (commit 4e8797fccc8e548a63ca2576d94de549eaadfa18) checks that the converted column meet simple rules defined for each bdf column. For example, cumulative quantities must be always positive, step quantities reset in line with the `step_id` column etc. Sanity checks our interpretation of cycler column names, so we have some evidence that we have assigned the correct labels. Will provide a mechanism for others to provide conflicting cases if they exist (Arbin case currently does not align with any of our definitions for capacity/energy)


tomjholland:
@jsimonclark currently a lot of tests are being skipped in the CI since only unit tests run and there are missing dependencies in the workflow. The dependency issue is easily fixed (probably the best option is to run with uv and specify --all-extras so that the CI environment is pinned to the uv.lock file like our dev environment). However many of the integration tests use the zenodo remote data, which I assume is why you weren't running them to begin with. These should be marked with the `pytest.mark.network` and should only fetch the remote file once (using the cache logic in `fetch`), but thought I would highlight here before updating the workflow in case you were guarding against something. You can run the tests yourself locally to evaluate.

jsimonclark:
@tomjholland good catch, you're right that CI only runs pytest tests/unit -q, so the whole tests/integration suite never actually runs in CI. I ran the full integration suite locally against this branch merged into current main fetching the real Zenodo data, which returned:

432 passed, 12 skipped, 13 deselected, 148 xfailed in 204s

the 148 xfailed are the intended assumed=False synonym markers. So the pipeline works on the real files, it just wasn't being exercised in CI.

No, I wasn't guarding against anything by excluding network tests. The marker split is fine, and your "fetch-once via the fetch cache" approach handles the cost issue. 

Please go ahead with the workflow change: uv + --all-extras pinned to uv.lock, with the Zenodo cases behind pytest.mark.network.

tomjholland:
@jsimonclark Okay sounds good. To clarify the plan, I can split up the workflow to minimise downloads. It will run: lint (ruff+mypy, on python 3.12) + build (build project and docs, on python 3.12) -> offline tests (pytest no network, full python matrix) -> network tests (pytest network, single python version). Therefore we stop the run before downloading anything if there is a failure outside of the network tests. Alternatively, we can use the workflow cache (https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching) between the no network and network tests to ensure the files are only downloaded once, allowing the full python version matrix for these tests too. What do you think?

jsimonclark:
> @jsimonclark Okay sounds good. To clarify the plan, I can split up the workflow to minimise downloads. It will run: lint (ruff+mypy, on python 3.12) + build (build project and docs, on python 3.12) -> offline tests (pytest no network, full python matrix) -> network tests (pytest network, single python version). Therefore we stop the run before downloading anything if there is a failure outside of the network tests. Alternatively, we can use the workflow cache (https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching) between the no network and network tests to ensure the files are only downloaded once, allowing the full python version matrix for these tests too. What do you think?

sounds good. Let's go with the cached full-matrix option. Matrix coverage on the network tests is worth it.

tomjholland:
@jsimonclark I've added the project-side updates necessary for the caching to work. Specifically:
- setting the cache dir by environment variable so the workflow can write to a reliable path in the cache
- enforcing that tests which access a url must be marked with pytest.mark.network, by blocking all network access to tests unless they have the network mark
- enforce that remote urls used in the notebooks are declared in a sidecar file so that they can be reliably included in the cache
- add the script to fetch the cache before the test matrix runs

The workflow prototype is here:
```yaml
name: CI

on:
  push:
    branches: [main, master]
  pull_request:

jobs:
  lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Install uv
        uses: astral-sh/setup-uv@v8.1.0
        with:
          python-version: "3.12"
      - name: Install dependencies
        run: uv sync --all-extras
      - name: Lint (ruff)
        run: uv run ruff check .
      - name: Type check (mypy)
        run: uv run mypy

  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Install uv
        uses: astral-sh/setup-uv@v8.1.0
        with:
          python-version: "3.12"
      - name: Install dependencies
        run: uv sync --all-extras
      - name: Check generated docs are in sync with ontology snapshot
        run: uv run python scripts/generate_docs.py --check
      - name: Build docs (link check skipped)
        run: uv run --all-extras sphinx-build -b html docs docs/_build/html
      - name: Build package and twine check
        run: |
          uv build
          uvx twine check dist/*

  test:
    needs: lint
    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
    env:
      MPLBACKEND: Agg
    steps:
      - uses: actions/checkout@v4
      - name: Install uv
        uses: astral-sh/setup-uv@v8.1.0
        with:
          python-version: ${{ matrix.python-version }}
      - name: Install dependencies
        run: uv sync --all-extras
      - name: Run tests (offline)
        run: uv run --all-extras pytest -q -m "not network"

  warm-cache:
    needs: test
    runs-on: ubuntu-latest
    env:
      BDF_CACHE_DIR: ${{ github.workspace }}/.bdf-cache
    outputs:
      cache-key: ${{ steps.key.outputs.k }}
    steps:
      - uses: actions/checkout@v4
      - name: Install uv
        uses: astral-sh/setup-uv@v8.1.0
        with:
          python-version: "3.12"
      - name: Install dependencies
        run: uv sync --all-extras
      - name: Compute content-addressed cache key (offline)
        id: key
        run: echo "k=bdf-data-$(uv run python scripts/warm_cache.py --emit-key)" >> "$GITHUB_OUTPUT"
      - name: Restore/save warmed cache
        uses: actions/cache@v4
        with:
          path: ${{ github.workspace }}/.bdf-cache
          key: ${{ steps.key.outputs.k }}
      - name: Warm cache
        run: uv run python scripts/warm_cache.py

  network:
    needs: warm-cache
    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
    env:
      MPLBACKEND: Agg
      BDF_CACHE_DIR: ${{ github.workspace }}/.bdf-cache
    steps:
      - uses: actions/checkout@v4
      - name: Install uv
        uses: astral-sh/setup-uv@v8.1.0
        with:
          python-version: ${{ matrix.python-version }}
      - name: Install dependencies
        run: uv sync --all-extras
      - name: Restore warmed cache (restore-only)
        uses: actions/cache/restore@v4
        with:
          path: ${{ github.workspace }}/.bdf-cache
          key: ${{ needs.warm-cache.outputs.cache-key }}
      - name: Run tests (network)
        run: uv run --all-extras pytest -q -m network
```
If there are no glaring issues you can see then I will push this to the PR branch and the workflow will run.

jsimonclark:
@tomjholland the proposed ci.yml looks good, please implement. But there is a conflict with main on pyproject.toml . Please keep your --disable-socket --ignore=... version.

PR #45, opened 2026-06-22
fix(units): don't read a quantity name as a unit in header parsing

jsimonclark:
`parse_from_header()` could misread a column **name** as a **unit**, silently corrupting cycle-count data. Pint defines `cycle = turn = revolution = 2π rad`, so a column named `Cycle` (Neware) or `z cycle` (Biologic) parsed as an angle, and normalization then scaled the dimensionless cycle count by 2*pi, every `cycle_count` value became `6.283185…`. `Step`/`Index` counts were spared only because Pint has no `step` unit.

## Root cause

In the snake-suffix branch, a token suffix was accepted as the unit even when it consumed the **entire** header (`base = " ".join(tokens[:-i]) or h` fell back to the full header as base but still returned the bogus unit), and angular unit-words were treated like any other unit.

```text
parse_from_header("Cycle")    -> ("Cycle", "cycle", "snake")   # before
parse_from_header("z cycle")  -> ("z",     "cycle", "snake")   # before
# normalize: convert_series(series, "cycle", "1")  ->  x 2π
```

## Fix

A cycler column header is a quantity name with an optional trailing unit, never a bare or angular unit. Two small, complementary guards in `parse_from_header`:

1. **require a non-empty base remainder** → a suffix can't consume the whole header (`Cycle`, `Bar`, `V`, `Ah`);
2. **reject angular unit tokens** (`cycle`/`turn`/`revolution`) → always a dimensionless count in this domain (covers `z cycle`, where the unit-word trails a prefix).

```text
parse_from_header("Cycle")    -> ("Cycle",   None, "none")  # after
parse_from_header("z cycle")  -> ("z cycle", None, "none")  # after
```

Counts now pass through unscaled. Explicit (`Voltage#V`, `Current (A)`) and genuine snake `name + unit` (`voltage_V`, `Power / W`) headers are unaffected.

## Testing

- New `tests/unit/test_parse_from_header.py` covers both collision shapes (whole-header and prefixed-angular) and confirms real units still parse.
- End-to-end: re-converting a Neware `.ndax` now yields `cycle_count = 1` (was `6.283185…`); voltage/current/capacity/step columns unchanged.
- Full unit suite + integration suite (real Neware/Biologic/Digatron fixtures) pass. *(The pre-existing `test_bundled_snapshot_is_up_to_date` failure on `main` is an unrelated stale bundled snapshot.)*

### DIFF
--- a/src/bdf/units/core.py
+++ b/src/bdf/units/core.py
@@ -49,6 +49,14 @@ _UNIT_ALIAS = {
     "per": "/",
 }
 
+# Tokens Pint recognises as angular units (1 cycle = 1 turn = 1 revolution = 2*pi
+# rad) but which, in a cycler column header, always denote a dimensionless count --
+# never an angle. Never accept them as a unit, or a cycle/turn count would be
+# silently scaled by 2*pi.
+_ANGULAR_UNIT_TOKENS = frozenset(
+    {"cycle", "cycles", "turn", "turns", "revolution", "revolutions"}
+)
+
 
 _SPLIT = re.compile(r"[ _\-]+")
 _HASH  = re.compile(r"^(?P<name>.+?)#(?P<unit>.+)$")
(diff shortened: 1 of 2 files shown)


jsimonclark:
@tomjholland , could you please review?

tomjholland:
@jsimonclark the fix looks sensible, but is this function still needed when we switch the backend to the new `table_normalizers.normalize()` method? I don't think this bug will be an issue for the header parser in the new code as it'll only use pint when the `{unit}` placeholder is specified in the synonym, but I can write a test to guard against it now we know this is how pint behaves. The PR I am working on now replaces the units module here with methods of the `spec` module, which performs the unit conversion directly using the instance methods of the ontology. Sorry I haven't had time to finalise and push that today, but will try to do so tomorrow so you can take a look.

jsimonclark:
The issue was resolved in subsequent PRs. Closing.

PR #49, opened 2026-06-24
feat(table_normalizers): add PyBaMM table normalizer

tomjholland:
Add a normalizer for a dataframe exported by `pybamm.Solution.get_data_dict()`. Currently converts minimal column set:
```python
{
  "Time [s]",
  "Current [A]",
  "Voltage [V]",
  "Discharge capacity [A.h]",
  "X-averaged cell temperature [C]",
  "X-averaged cell temperature [K]",
}
```


jsimonclark:
Could you give a description for this PR? A quick check shows that:

**The temperature header doesn't exist in PyBaMM.** There is no `"X-averaged cell temperature [C]"` variable (checked pybamm 26.6.2 ), so temperature is silently dropped for every real PyBaMM export.

tomjholland:
@jsimonclark I'm not sure why it wasn't appearing on your installation, but `"X-averaged cell temperature [C]"` is a valid argument to `pybamm.Solution.get_data_dict()`, as is the same quantity in Kelvin. I've made the Synonym unit-agnostic and added an integration test that exports a real PyBaMM solution (with PyBaMM as an optional dependency, and the test skips if not installed).

PR #50, opened 2026-06-24
feat(io): add tz parameter for naive datetime parsing

tomjholland:
## Summary
Adds a `tz` parameter to `read`/`normalize`/`TableNormalizer.normalize` so vendor formats without an embedded UTC offset (Arbin, Maccor, Neware, Novonix, LANDT) can be localized to an explicit IANA timezone instead of always being treated as UTC.

## Features
- New `tz: str = "UTC"` kwarg threaded through `bdf.read`, `bdf.normalize`, and `TableNormalizer.normalize`, localizing naive `unix_time_second` formats to the given IANA zone ([7fdc738](https://github.com/tomjholland/battery-data-format/commit/7fdc738))
  - Formats with an embedded offset (e.g. Digatron's `%:z`) ignore `tz` entirely — the embedded offset is authoritative
  - Elapsed-time columns (`test_time_second`/`step_time_second`) are unaffected since the offset cancels out
  - Invalid `tz` values raise `ValueError` up front, before any lazy collect
  - Leaving `tz` at its default emits a `UserWarning` when a naive format is in play, so the UTC assumption is never silent

## Docs
- Document the `tz` parameter and default-UTC warning behavior in the user guide ([7fdc738](https://github.com/tomjholland/battery-data-format/commit/7fdc738))

## Tests
- Add coverage for naive-format localization, tz-aware formats ignoring `tz`, the default-UTC warning, invalid `tz` validation, and elapsed-time invariance ([7fdc738](https://github.com/tomjholland/battery-data-format/commit/7fdc738))

## Refactoring
- Rename `rc` to `resolved_column` in `TableNormalizer` for clarity ([b25d783](https://github.com/tomjholland/battery-data-format/commit/b25d783))


jsimonclark:
Good feature. One thing:

**Daylight savings time transitions crash the read.** `replace_time_zone(tz)` defaults to `ambiguous='raise'` / `non_existent='raise'`.  In the event of a DST changeover, the fall-back hour exists twice (ambiguous), and if the logging PC doesn't adjust its clock, it also writes times inside the spring-forward gap (non-existent). One such row makes the whole read raise `ComputeError` , and with `lazy=True` it surfaces at `.collect()`, with no mention of `tz`. Repro on polars 1.41.2: `2024-03-10 02:30:00` with `tz="America/New_York"` → "non-existent"; `2024-11-03 01:30:00` → "ambiguous".

Suggest passing an explicit strategy, e.g. `ambiguous='earliest'` (correct for monotonically increasing test time) and `non_existent='null'`, or catching the error and re-raising with a hint to use a fixed-offset zone (`Etc/GMT+5`) for loggers that don't observe DST.

tomjholland:
@jsimonclark I think this is resolved. I went with the explicit strategy as catching the error becomes tricky if the LazyFrame is not resolved within bdf

Issue #52, opened 2026-07-11
Reference file non-compliance (step_index holds Step IDs; cumulative/net columns swapped) + add derived-column checks to validate()


pghege:
## Summary

While reviewing the reference example `docs/examples/reference/FZJ__INR21700__20250606__HPPC__25degC__Digatron.bdf.csv` (47,927 rows), we found two genuine non-compliances with the ontology definitions, plus a related gap: `bdf.validate()` does not currently check any of these, so the bad data passes validation. Reported by an external adopter (FZJ/Digatron) trying to confirm their converter output is compliant.

All findings below were verified numerically against the file and the definitions in `src/bdf/data/bdf-ontology-snapshot.ttl`.

## Finding 1 — `step_index` column contains Step IDs (non-compliant)

Ontology `step_index` ("Step Index / 1") is a **1-based within-step point counter** that *"resets to 1 at the start of each new step and increments by 1 for each recorded data point… an instrument column named 'Step Index' (e.g. Arbin Step_Index) maps to `step_id`, not to this property."*

In the file, `step_index` holds values `{4,5,6,8,9,…,9999}`, **never equals 1**, and sits in 58 flat plateaus — i.e. it is the **Digatron program Step** (`step_id`), placed in the wrong column. `step_id` is a documented equivalent of "Digatron Step".

**Fix:** map Digatron Step → `step_id`; derive a true `step_index` (1,2,3… resetting per step) if that column is to be present.

## Finding 2 — `cumulative_*` and `net_*` columns are swapped / wrong-signed (non-compliant)

Ontology: `cumulative_capacity_ah = charging_capacity_ah + discharging_capacity_ah` (**always monotonically non-decreasing**); `net_capacity_ah = charging_capacity_ah − discharging_capacity_ah` (may be negative). Same pattern for energy.

Measured (100% of rows):

| File column | Actually equals | = ontology definition of |
|---|---|---|
| `cumulative_capacity_ah` | charging − discharging | `net_capacity_ah` |
| `cumulative_energy_wh` | charging − discharging | `net_energy_wh` |

The file's `cumulative_capacity_ah` ends **negative (−2.12) and is non-monotonic**, which is impossible for a true cumulative. `charging_*`/`discharging_*` primitives are themselves correct and monotonic, so the fix is a relabel/sign correction of the cumulative & net columns.

## Finding 3 — `cycle_count` is not monotonic (minor)

[...]

PR #53, opened 2026-07-11
validate: check ontology-defined derived-column consistency

pghege:
## What

Adds warning-level **derived-column consistency checks** to `validate_df()`, so that data which is structurally valid but internally inconsistent with the ontology's derived-quantity definitions is surfaced instead of passing silently.

Closes the "root gap" identified in #52.

## Why

An external adopter (FZJ / Digatron) asked for a reliable, machine-checkable definition of how the derived columns relate to the primitives, so they can confirm their converter output is compliant. Today `validate_df()` only checks (a) required-column presence and (b) `Test Time / s` monotonicity — none of the algebra the ontology defines is enforced.

## Checks added

All derived from `prov:wasDerivedFrom` / `skos:definition` in `bdf-ontology-snapshot.ttl`. Columns are resolved from preferred labels, machine-readable notations, **or** vendor synonyms, so checks work on any accepted header style.

- `cumulative_{capacity_ah,energy_wh}` == `charging_* + discharging_*`, and monotonically non-decreasing
- `net_{capacity_ah,energy_wh}` == `charging_* − discharging_*`
- `charging/discharging_{capacity_ah,energy_wh}` monotonically non-decreasing
- `cycle_count` non-negative, integer-valued, monotonically non-decreasing
- `step_index` is a 1-based within-step point counter (resets to 1, else +1) — flags the common case where a program **Step ID** (Arbin Step_Index / Digatron Step / Neware Step_ID) has been placed in `step_index`

Findings surface as `RuntimeWarning`s and in the `report=True` printout. Derived columns remain **optional**, so this never turns into a hard validation failure.

## Scope / minimality

Two files, additive only, **zero deletions**, no changes to existing logic:

```
 src/bdf/validate.py         | 157 +++++++++++++++++
 tests/unit/test_validate.py |  56 ++++++++
```

## Validation

- `pytest tests/unit` → **552 passed, 13 skipped**
- `ruff` clean; `mypy` clean (CI-equivalent isolated env: "no issues found in 21 source files")
- Ran against **all 10** reference example files — full results in #52. Summary: the 4 DLR/Basytec files are clean; FZJ, the 3 SINTEF Neware/Landt files, and the 2 BioLogic files each trip one or more checks (mislabeled `step_index`, swapped `cumulative`/`net`, a `cycle_count` column equal to a constant 2π, and small monotonicity glitches).

Refs #52.

### DIFF
--- a/src/bdf/validate.py
+++ b/src/bdf/validate.py
@@ -106,12 +106,8 @@ def _check_derived(df: pd.DataFrame) -> Dict[str, Any]:
         n_bad = int(mismatch.sum())
         if n_bad:
             worst = float(np.abs(got[mismatch] - exp[mismatch]).max())
-            issues.append(
-                f"'{target}' != {a} {op} {b} in {n_bad}/{len(df)} rows (worst |Δ| = {worst:.4g})."
-            )
-            details.append(
-                {"check": "identity", "column": target, "violations": n_bad, "worst_abs_diff": worst}
-            )
+            issues.append(f"'{target}' != {a} {op} {b} in {n_bad}/{len(df)} rows (worst |Δ| = {worst:.4g}).")
+            details.append({"check": "identity", "column": target, "violations": n_bad, "worst_abs_diff": worst})
 
     # 2) monotonic non-decreasing quantities
     for name in _MONOTONIC_NONDECREASING:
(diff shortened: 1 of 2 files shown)


jsimonclark:
Pushed a formatting-only fixup (5132139) so pre-commit goes green. This repo's ruff config uses `line-length = 120` (ruff.toml), so `ruff format` collapses several of the wrapped calls onto single lines; easy to miss if your editor formats at the default 88. No functional change; all your tests pass unchanged.

jsimonclark:
Reviewed the implementation and exercised the checks against real files: they correctly flag every non-compliance from the #52 audit on the current reference set, and pass cleanly on files regenerated from raw with the current reader. Design choices are right: warning-level (derived columns are optional, but must be consistent when present), header-style agnostic via label/notation/synonym resolution, and sensible tolerances.

Merging. The review for #57 to follow once the reference-data refresh lands (its known-bad allowlist is keyed to files that refresh fixes).

Issue #56, opened 2026-07-11
Reference file: Neware G20M7 cycle_count is a constant 2π (+ minor cycle_count/monotonicity glitches)


pghege:
Follow-up from #52 (checks added in #53).

**`SINTEF__G20M7-202512-Gru6mV__20251228__C30__25degC__Neware.bdf.csv`**: the entire `cycle_count` column is the constant `6.283185307179586` — i.e. **2π**. `cycle_count` must be a non-negative **integer**, monotonically non-decreasing. This looks like a converter wiring bug (an angle/constant landed in `cycle_count`).

Two smaller `cycle_count` items found in the same audit:
- `FZJ__INR21700__20250606__HPPC__25degC__Digatron`: `cycle_count` is not monotonically non-decreasing (3 backward steps).
- Several SINTEF Neware/Landt/BioLogic files have 1–3 tiny non-monotonic drops in `charging/discharging_capacity_ah` / `_energy_wh` — likely minor data glitches, worth a look during the cleanup.

All flagged by `validate_df` (#53).

Issue #55, opened 2026-07-11
Reference files: cumulative/net capacity & energy columns swapped (FZJ, BioLogic)


pghege:
Follow-up from #52 (checks added in #53). The `cumulative_*` and `net_*` columns are swapped / wrong-signed in some reference files: the "cumulative" column actually holds net values.

Ontology definitions:
- `cumulative_capacity_ah = charging + discharging` (always monotonically non-decreasing)
- `net_capacity_ah = charging − discharging` (may be negative); same for energy.

Verified:

| File | Observation |
|---|---|
| `FZJ__INR21700__20250606__HPPC__25degC__Digatron` | `cumulative_capacity_ah` == charging−discharging (net) in 100% of rows; ends negative (−2.12) and non-monotonic. Same for `cumulative_energy_wh`. |
| `SINTEF__NaCR32140-MP10-04__2025-08-25__GITT_0p05C_25degC__BioLogic` | `cumulative_energy_wh` == net (100%); capacity identity also broken |
| `SINTEF__NaCR32140-MP10-04__2025-08-25__CCCV_0p02C_25degC__BioLogic__Outlier_Bug` | cumulative identities broken (this file is already flagged as an outlier bug) |

For the BioLogic files the capacity side is murkier than a clean swap — likely tied to BioLogic EC-Lab reporting **per-step** Q (which resets each step), whereas `charging/discharging_capacity_ah` must accumulate from test start (the ontology calls this out explicitly). Worth confirming the converter isn't feeding per-step Q into the cumulative primitives.

**Fix:** correct the column mapping/sign so `cumulative_* = charging + discharging` (non-decreasing) and `net_* = charging − discharging`. Flagged by `validate_df` (#53).

Issue #54, opened 2026-07-11
Reference files: Step ID placed in step_index column (FZJ, SINTEF Landt/Neware)


pghege:
Follow-up from #52 (checks added in #53). Several reference files place a program **Step ID** (Arbin Step_Index / Digatron Step / Neware Step_ID) into the `step_index` column. Per the ontology, `step_index` is a 1-based *within-step point counter* (resets to 1, +1 per point); the program step number belongs in `step_id`.

Affected files (verified):

| File | `step_index` content |
|---|---|
| `FZJ__INR21700__20250606__HPPC__25degC__Digatron` | values `{4,5,6,…,9999}`, never 1 — Digatron Step |
| `SINTEF__LiGrR2032__2024-04-30__25degC__Landt` | 3 distinct values, long plateaus |
| `SINTEF__G20M7-202512-Gru6mV__20251228__C30__25degC__Neware` | 6 distinct values, long plateaus |
| `SINTEF__SLPBA842124HV__2024-10-23__Rate_25degC__Neware__Time_Bug` | 20 distinct values, long plateaus |

The DLR/Basytec reference files are unaffected.

**Fix:** map the instrument step column to `step_id`; if `step_index` is to be present, derive it (1,2,3… reset per step). `validate_df` (#53) now flags this.

PR #57, opened 2026-07-11
Prevent malformed reference data at the source (vendor-header disambiguation + generator/CI gates)

pghege:
## What

Root-cause follow-ups to #53 (detection) so the class of corruption audited in #52 can't recur. Now rebased onto main (#53 merged, plus #63/#64/#67 follow-ups).

Key correction from the #52 investigation: **bdf's own ingest is correct.** Running the one in-repo raw file (`data/…Landt.csv`) through `bdf.read()` yields `Step ID` (not a bogus `step_index`) and clean derived columns. The corruption lives in reference `.bdf.csv` files as they were historically produced/uploaded, which the generator downloads. So the fixes here are about **prevention and disambiguation**, not fixing bdf's translator.

## Changes

**1. Vendor header disambiguation (`spec.py`)** — the generic `base_synonym_index` mapped `step index → step_index`, contradicting the ontology's own rule that a raw "Step Index" (Arbin) / "Step" (Digatron) / "Ns" (BioLogic) is a program step **identifier** (`step_id`). Added `_VENDOR_SYNONYM_OVERRIDES`:
- `step index`, `step`, `ns` → `step_id`
- `cycle index` → `cycle_count` (was unmapped)
- canonical `Step Index / 1` (slug `step-index-1`) still → `step_index` via the exact-label path, so BDF round-trips are unaffected.

**2. Generator gate (`generate_zenodo_reference_bdf.py`)** — it ran `bdf.read(..., validate=False)` and never checked its output. Now every generated file is validated against the derived-column rules; issues are recorded in `manifest.json`, warned per-file, and `--strict` fails the run on any inconsistency.

**3. CI gate (`test_reference_data_valid.py`)** — validates every `docs/examples/reference/*.bdf.csv`. Files not in `KNOWN_NONCOMPLIANT` must be clean (blocks new bad data); files in it (tracked in #54–#56) must still be non-compliant (so upstream fixes force the allowlist to shrink). The 4 DLR/Basytec files pass clean today.

## Not included (needs a decision)

**Regenerating the Landt reference from raw** via `bdf.read` produces a *clean* file but a wholesale reshape — 7→16 columns, snake_case→canonical labels, and it drops the cumulative `charging/discharging_capacity_ah` columns in favour of step-level ones. That's a semantic replacement, not a patch, so I left it out pending a call on whether reference files should carry cumulative or step-level capacity.

## Validation

- `pytest tests/unit tests/integration/test_reference_data_valid.py tests/integration/test_table_normalizers.py` → **563 passed, 14 skipped**
- `ruff` clean; `mypy` clean (CI-equivalent isolated env)

Refs #52, #54, #55, #56.


pghege:
Rebased onto current main (reconciled with #63/#64/#67 — the vendor-synonym overrides compose cleanly with the expanded unit regex and the isReplacedBy redirects). Fixed the ruff-format failure that was blocking the test matrix; all checks are green and the PR is mergeable. Ready whenever it fits the pre-0.2.0 sequence.

pghege:
Rebased onto current main (153 commits: the API freeze, polars ports, module rearrangement). Changes in the port, for reviewers:

- The generator gate now uses the frozen API: `bdf.save(df, path, labels=...)` and `bdf.validate_df` on polars directly, no pandas round-trip.
- The stale `from bdf.validate import ...` module path (removed in #94) is gone.
- Ontology 1.3.0 sharpened this PR's premise: the deprecated `step_index` synonym now redirects the `step-index` slug to `step_record_index` via isReplacedBy, so an unmapped vendor "Step Index" header would be silently relabeled as a per-step record counter. The override table and its tests are updated to pin the current names (`step-record-index-1` -> `step_record_index` stays intact).

732 unit tests + the reference gates pass locally. Still sequenced to merge alongside the regenerated reference bundle.

PR #60, opened 2026-07-14
Add Arbin RES support

Abbta:
### Summary
Adds support for Arbin `.res` files using `polars-access-mdbtools` to parse MDB/RES files. Note that `polars-access-mdbtools` requires MDB Tools.

### Added
- Parser for MDB files
- Normalizer for Arbin `.res` which does not use CSV style column names
- Unit tests for parsing `.res` files, also added `.res` parser to the integration tests

### Comments
- For installation of MDB Tools, the installation guide of `polars-access-mdbtools` is linked to in README.
- When parsing `.res` files, `polars-access-mdbtools` appears to return rows out of data-point order. This is solved by forcing a sort by column `Data_Point` in `MDBParser`. It is not clear how `polars-access-mdbtools` would interact with other MDB type files.
- The `.res` files use OADate; the normalizer converts this to Unix seconds.
- For the integration tests I temporarily added one of my own `.res` files, please let me know how you would like the test data.
- The integration test case records 5 expected validity xfails.

### Test results
Result for the full test suite: 985 passed, 12 skipped, 153 xfailed, 36 warnings


jsimonclark:
Thanks for contributing @Abbta . Solid plugin, and the mappings match what we have on the ontology side. Three changes needed before merge:

1. Please remove `tests/data/arbin/sample.res` (5.3 MB) from the branch history (rebase/force-push ; a deletion commit isn't enough). We keep small files (< ca. 500 kb) in the repo for quick reference, but larger files live in our Zenodo test-data record (https://zenodo.org/records/21337233), not git. I can upload it there credited to you. Just confirm you can share it under CC-BY 4.0 and give a one-line description (cell/chemistry, test type). Then we point the tests at the download URL like the other integration cases.
2. CI never runs the RES tests. They all skip with "MDB Tools not installed" (33 skipped in the green run). Add `sudo apt-get install -y mdbtools` to the test jobs.
3. Set `polars-access-mdbtools>=0.1.2` in the arbin extra. Unconstrained, old environments resolve 0.0.2, which has an incompatible API.

Two quick questions while you're here: is Internal_Resistance a DC pulse measurement, and are AC_Impedance/ACI_Phase_Angle |Z| and phase at the ACIM excitation frequency? Your confirmation settles what we've only inferred from data.

On the accumulator resets you flagged in known_validity_bugs: agreed, and we're discussing schedule-scoped ontology terms for exactly this. We'll remap both Arbin dialects in a follow-up after the next ontology release. Keep your mapping as is.

@tomjholland @pghege , do you agree? anything to add?

Abbta:
Thanks for the feedback, it should be addressed now. Integration tests currently point at nonexistent file. I will get back to you later with the file description and your questions on the resistance and impedance.

g-kimbell:
Hi @Abbta, thank you for your contribution! Reading res files would definitely be useful.

For Windows users, installing mdb-tools is a bit painful. We could also try `pyodbc` - it can use Microsoft's access engine which many users probably have already:
```python
driver = next((d for d in pyodbc.drivers() if "Microsoft Access Driver" in d), None)
conn = pyodbc.connect(rf"DRIVER={{{driver}}};DBQ={filepath.resolve()};")
df = pl.read_database("SELECT * FROM Channel_Normal_Table", conn).sort(by = "Data_Point")
```

I think users are going to struggle a lot with the weird extra dependencies and different ways to install on different operating systems, so we need checks with really clear error messages. `polars-access-mdbtools` just spits out a generic "FileNotFoundError" which will trip people up.

I'd suggest:
```python

if sys.platform == "win32" and _has_access_driver():  # pyodbc.drivers() check above
    return _read_with_pyodbc(filepath, table)
if _has_mdbtools():  # e.g. import + shutil.which("mdb-export")
    return _read_with_mdbtools(filepath, table)
raise error with useful os-specific instructions
```

I'd also consider pinning the exact `polars-access-mdbtools==0.1.2`, as its a 0.1.x on a small repo, they may introduce breaking changes any time.

Abbta:
I addressed the feedback from @g-kimbell; `pyodbc` works well for MDB files. As a test file you can use
[this file](https://github.com/user-attachments/files/30371584/Anwesa-20231222-PW-Na-ECSEC.zip). I consent to it being shared under CC-BY 4.0. It is a constant current test on a prussian white sodium ion cell.

As for the questions on the Arbin measurments by @jsimonclark I can confirm that the answer is yes to both questions (Arbin documentation on [internal resistance measurements](https://www.arbin.com/how-to-perform-dc-internal-resistance-measurement-with-arbin.html?) and [impedance measurements](https://www.arbin.com/integrated-electrochemical-impedance-spectroscopy-eis-with-arbin-charge-discharge-channels.html)).

pghege (commented):
Thanks for this, and sorry it sat waiting so long. The parser itself is in good shape: the dual backend with clear install guidance, the mocked unit tests, the CI mdbtools install, and the extras with platform markers are all done carefully. Since it inherits read() from the base class it also picked up the new #76 API without changes.

One substantive thing before it can merge, and it's a recent development rather than something you missed:

**Capacity and energy columns should map to the new schedule_ terms.** We learned from Arbin directly (their MITS team) that Charge_Capacity, Discharge_Capacity, Charge_Energy and Discharge_Energy reset at operator-defined points in the test schedule, sometimes per step, per cycle, irregularly, or never. That means they don't match `charging_capacity_ah` and friends, which are defined as never resetting. Ontology 1.3.0 added `schedule_charging_capacity_ah`, `schedule_discharging_capacity_ah`, `schedule_charging_energy_wh` and `schedule_discharging_energy_wh` for exactly these columns, and the fields exist on main since #79. The .csv and .xlsx Arbin normalizers are being remapped the same way as part of the release.

Two smaller things:

- The DateTime conversion (Access day-fraction to unix seconds) assumes the recorded time is UTC. Access datetimes are naive local wall-clock, and our other naive-datetime paths go through the `tz` parameter. At minimum worth a comment documenting the UTC assumption; routing through the datetime handling would be better if practical.
- polars-access-mdbtools needs Python 3.11+, so on 3.10 the extra installs nothing and users only find out at runtime. A line in the README install note would save some confusion.

You've allowed maintainer edits on this branch, so if you'd rather not chase the schedule_ remap yourself I'm happy to push that change here, just say the word.

pghege:
Went ahead and helped with both items rather than leaving you homework: merged main into this branch (brings the new read/scan API and the ontology 1.3.0 snapshot) and applied the schedule_* remap to the RES normalizer, matching what's being done for the csv and xlsx normalizers in #82. Full unit suite passes. Remaining from the review: the DateTime UTC assumption comment and the Python 3.11 note in the README, both small. Happy to approve once those are in, or add them too if you prefer.

g-kimbell:
This works nicely in my tests!

Abbta:
Great! Thank you for the feedback and also the help. I can have a look at those two small issues later this week but if you've got time @pghege, be my guest to fix them yourself.

jsimonclark:
> I addressed the feedback from @g-kimbell; `pyodbc` works well for MDB files. As a test file you can use [this file](https://github.com/user-attachments/files/30371584/Anwesa-20231222-PW-Na-ECSEC.zip). I consent to it being shared under CC-BY 4.0. It is a constant current test on a prussian white sodium ion cell.
> 
> As for the questions on the Arbin measurments by @jsimonclark I can confirm that the answer is yes to both questions (Arbin documentation on [internal resistance measurements](https://www.arbin.com/how-to-perform-dc-internal-resistance-measurement-with-arbin.html?) and [impedance measurements](https://www.arbin.com/integrated-electrochemical-impedance-spectroscopy-eis-with-arbin-charge-discharge-channels.html)).

@Abbta, how would you like be cited for CC-BY-4.0 on the .res file? Do you have an ORCID?

pghege (approved):
All review items resolved (the last two I pushed directly per your invitation). Thanks for the contribution and the test file, great to have .res support in.

Abbta:
> > I addressed the feedback from @g-kimbell; `pyodbc` works well for MDB files. As a test file you can use [this file](https://github.com/user-attachments/files/30371584/Anwesa-20231222-PW-Na-ECSEC.zip). I consent to it being shared under CC-BY 4.0. It is a constant current test on a prussian white sodium ion cell.
> > As for the questions on the Arbin measurments by @jsimonclark I can confirm that the answer is yes to both questions (Arbin documentation on [internal resistance measurements](https://www.arbin.com/how-to-perform-dc-internal-resistance-measurement-with-arbin.html?) and [impedance measurements](https://www.arbin.com/integrated-electrochemical-impedance-spectroscopy-eis-with-arbin-charge-discharge-channels.html)).
> 
> @Abbta, how would you like be cited for CC-BY-4.0 on the .res file? Do you have an ORCID?

You can site me by my name Abbe Tanndal, however the original creator of the file is Anwesa Mukherjee with ORCID [0009-0006-3470-1907](https://orcid.org/0009-0006-3470-1907).

jsimonclark:
The Zenodo record with test datasets has been updated to include the contribution from @Abbta : https://doi.org/10.5281/zenodo.16994937

PR #61, opened 2026-07-14
Add Biologic .mpr parsing with yadg

### LINKED ISSUE #15: Read Biologic raw mpr using `yadg`

g-kimbell:
Adds Biologic binary .mpr parsing using yadg. Second PR attempt - #18 was for the old read methods. Closes #15.

Adds an MPR `TableParser`, updates the biologic `TableNormalizer`, adds test data and test cases

⚠️ This PR also adds `reverse_sign` as an option in `Syn`
This lets you reverse the sign of columns independently of the unit conversion
Useful here because biologic records `-Im(Z)/{unit}`
This means still get unit parsing and other synonyms, which isn't possible when defining a `ResolvedColumn` like we did with the discharge-positive currents in the PyBaMM normalizer


g-kimbell:
@jsimonclark I have some files to cover the gaps in the tests, how do want test data uploaded?

I have some small datasets (<100 kb) that I could put in tests/data.

Or do I need to put everything on Zenodo?

jsimonclark:
datasets < 100 kb can go in the repo. Otherwise we put them in Zenodo. Just be sure it's ok to share them under the respective licenses.

g-kimbell:
@jsimonclark more weird Python 3.10 bugs, but finally got green ticks - ready for review

This depends on #67 so I'll need to do a rebase if you squash that, let me know.

@tomjholland I added `reverse_sign` to `Syn` to negate columns while keeping unit parsing and the ability to have other synonyms. Here its used for Biologic's negative imaginary impedance, it might also be useful for PyBaMM table normalizer if we ever need unit parsing. Let me know if there's anther way you'd prefer to do this.

tomjholland:
> @jsimonclark more weird Python 3.10 bugs, but finally got green ticks - ready for review
> 
> 
> 
> This depends on #67 so I'll need to do a rebase if you squash that, let me know.
> 
> 
> 
> @tomjholland I added `reverse_sign` to `Syn` to negate columns while keeping unit parsing and the ability to have other synonyms. Here its used for Biologic's negative imaginary impedance, it might also be useful for PyBaMM table normalizer if we ever need unit parsing. Let me know if there's anther way you'd prefer to do this.
> 
> 

Yep I'm happy with that, solves the problem simply

jsimonclark:
Thanks @g-kimbell Looks good overall! Just a few changes before merge and one open question for @pghege 

## Issues

1. **Division-by-zero in the synthesized current.** In `_fix_missing_columns`, `I/mA = dq/Δt` with `Δt` from `uts.diff()` , duplicate timestamps (which BioLogic files do produce) give `Δt = 0` creating infinite current. Please guard that case (null or 0 where `Δt == 0`).
2. **Cycle indexing.** Synthesized `cycle number = half cycle // 2 + 1` is 1-indexed, but BioLogic's native `cycle number` is 0-indexed. The same test would report different cycle numbers depending on which column the file happens to contain. 
3. **Provenance.** Please confirm the four `.mpr` files are your own/Empa measurements and shareable under the repo license?
4. Rename the extra from `yadg` to `mpr` ; other extras are named by format (`excel`, `mat`), not by backing library.

## Suggestions

- A short comment where `_err` columns are dropped, since that filtering is invisible to the normalizer's unmapped-column accounting.

## On the yadg dependency  
yadg is GPL-3.0, which would be this project's first copyleft-linked dependency (fastnda is BSD-3, the rest are permissive). The way you've done it (optional extra, runtime import, nothing vendored) keeps bdf's Apache-2.0 core clean, and only affects downstream users who redistribute bundled environments. 

Whether we adopt that as project policy is a governance question rather than a review one, so I'm deferring that decision to @pghege . If it's accepted, we'll add a note to the README/docs that the extra links a GPL-3.0 library.

## Notes
This branch contains the #67 commits, which are now on main. It merges clean as-is, but a rebase would tidy the commit list.

g-kimbell:
Rebased and fixed the issues you raised. Also pinned yadg <7.0 as it'll release soon with breaking changes I'd need to test.

All the example files are from our lab and free to share under the repo license.

pghege (commented):
Reviewed the full diff. This is in good shape, and it's nice to see the EIS terms getting their first real exercise. The parser inherits read() from the base class, so it picked up the #76 API changes for free, and the merge state is clean. Three things before I approve:

1. Docstring says "Wraps fastnda for Biologic .mpr binary files", should be yadg.

2. `_fix_missing_columns` synthesizes data in two places worth a second look. When current is absent it creates an all-zero `I/mA` column, and when only `dq` exists it derives current as dQ/dt. Both are reasonable, but the file then contains columns that were computed, not measured, with nothing recording that. Suggest at least a UserWarning when a column is synthesized, so downstream users know. Same question for `half cycle // 2 + 1`: synthesized cycle numbers start at 1 while real BioLogic cycle numbers start at 0, so the same cell looks different depending on which columns the export happened to include.

3. `reverse_sign` is a clean mechanism and the `-Im(Z)` use is exactly right against the term definition. Could you add a small unit test for it in test_table_normalizers so the behavior is pinned there too, not only via the integration case?

The yadg dependency is already covered by the existing `mpr` extra, so nothing needed there.

g-kimbell:
Hi @pghege, thanks for checking over this

- Rebased and tested on latest main
- Fixed docstrings
- `_fix_missing_columns` now warns, added unit tests to check this and check logic
- Reconstructed `cycle_count` is now 0-indexed, not 1-indexed, like in mpt files. I did make this commit before but it got lost somewhere, sorry.
- Added tests for `reverse_sign`

PeterKraus:
Very happy to see `yadg` is getting used. A couple of comments / suggestions:

- we also have a `mpt` parser; I know you already have your own normalizer, but just in case you would like to reduce your own maintenance burden, you could just treat `mpr` and `mpt` files the same and offload them to `yadg`.
- @jsimonclark: if you have a file with duplicate timestamps, I'd be very happy to add it to our test suite. I didn't know this was a possibility in those files.
- @g-kimbell: the correct way to figure out which columns you can ignore (as you currently do for `*_std_err`) would be to look through the `ancillary_variables` attributes. This should make your harness compatible with `yadg-7.0`, too, where we started using `*_uncertainty` instead of `*_std_err` for clarity reasons.

g-kimbell:
Thanks for the tip @PeterKraus 
Using `ancillary_variables` to detect uncertainty vars now
All tests passing on my machine for both yadg 6.2.4 and 7.0, so I've relaxed the version pinning in pyproject toml

pghege (approved):
All review items addressed, and the upstream input from Peter made this better than the version I reviewed. The 0-indexed rebuilt cycle numbers resolve the consistency point too. Nice work.

Issue #65, opened 2026-07-14
Digatron CSV export: time column labeled seconds contains milliseconds (FZJ HPPC reference file)


jsimonclark:
## What  
The Digatron export in the reference set (`FZJ__INR21700__20250606__HPPC__25degC__Digatron.csv`, Zenodo 10.5281/zenodo.18986774) declares the test-time column in seconds in its header, but the values are actually milliseconds.

## Impact  
`bdf.read` correctly trusts the declared unit, so the converted artifact's timeline is uniformly wrong by 1000×. All internal validation passes, the error is self-consistent, which is exactly what makes it dangerous: nothing in the file alone reveals it.

## How it was found  
Cross-checking the vendor test-time increments against the independently recorded wall-clock (`DateTime`) increments; the two disagree by a factor of ~1000 on this file and agree on every other reference file.

## Plan  
1. The reference-datasets scorecards gain this wall-clock cross-check, so the FZJ scorecard reports the defect openly instead of passing silently (reference datasets PR).
2. A time-reconciliation feature (design under discussion) will detect the mismatch on read, repair the scale, and record what it did in metadata, after which the FZJ artifact is regenerated and its scorecard goes green.

This issue closes when step 2 lands and the regenerated artifact validates.

Issue #58, opened 2026-07-11
Tracking: reference dataset (Zenodo 18214281) compliance audit — full scope, provenance, remediation


pghege:
## Summary

The reference datasets under `docs/examples/reference/*.bdf.csv` are internally inconsistent with the BDF ontology's own column definitions. An external adopter (FZJ) first flagged it (#52); auditing all 10 files shows the problem is **systematic across contributors, not isolated**. This issue consolidates the full scope, provenance, root cause, and remediation status so the other maintainers and data contributors can be looped in.

**Key finding:** `bdf`'s own ingest is **correct** — running raw source through `bdf.read()` produces conformant BDF. The corruption is in the **already-converted BDF that was uploaded to Zenodo** and which this repo downloads verbatim.

## Provenance

Reference data is downloaded from Zenodo record **[10.5281/zenodo.18214281](https://doi.org/10.5281/zenodo.18214281)** — *"Battery Datasets for Testing Data Pipelines and Workflow Automation"* (CC-BY-4.0, published 2026-01-11) via `scripts/generate_zenodo_reference_bdf.py`.

Record creators (per institution):
| Institution | Creators | Datasets |
|---|---|---|
| SINTEF | Simon Clark, Julian Tolchard | Landt, Neware ×2, BioLogic ×2 |
| DLR (German Aerospace Center) | Dennis Kopljar, Christina Schmitt | Basytec ×4 |
| Forschungszentrum Jülich (FZJ) | Elias Barbers | Digatron HPPC |

(The git author on the committed `.bdf.csv` files reflects who ran the download notebook, not the data donor.)

## Full audit (all 10 reference files)

Verified numerically against the ontology definitions; detection added in #53.

| File | Donor | Status |
|---|---|---|
| DLR…Basytec (GITT, POCV, GITT, POCV) ×4 | DLR | ✅ **clean** — conformant reference examples |
| FZJ…HPPC…Digatron | FZJ | ❌ step_id in `step_index`; `cumulative`/`net` swapped (cap+energy); `cycle_count` non-monotonic |
| SINTEF…Landt | SINTEF | ❌ step_id in `step_index`; minor monotonic glitches |
| SINTEF…G20M7…Neware | SINTEF | ❌ step_id in `step_index`; **`cycle_count` is a constant `6.283185…` (= 2π)**; monotonic glitches |
[...]

Issue #51, opened 2026-07-01
Interoperability plans + request for a stable spec and compliant reference data


simontaurus:
## Context

We are building interoperability with the battery-data-format (BDF) in
[opensemantic.batteries](https://github.com/OpenSemanticWorld-Packages/opensemantic.batteries-python)
- a typed, unit-aware battery cycling data model built on QUDT-derived
quantities. We would like our datasets to map cleanly to and from BDF.

## Request

Before we commit to a fixed BDF mapping on our side, it would help a lot to have:

1. A **stable, versioned spec** for the column keys (and their units).
2. A set of **compliant, bug-free reference datasets** to validate against.

## Discrepancies we ran into

While prototyping the mapping we found the naming and reference data not yet
fully consistent:

- The machine-readable names in the ontology/docs (e.g. `test_time_second`,
  `voltage_volt`, `current_ampere`) do not match the column headers in the
  repo's own data, which use `test_time_s`, `voltage_V`, `current_A`.
- Two representations coexist (the Preferred Label `Test Time / s` vs the
  machine-readable alias) without a single normative form for software to key on.
- `data/` and `examples/reference/*.bdf.csv` appear to use different conventions.
- Several reference files are explicitly marked as containing issues
  (`..._Time_Bug.bdf.csv`, `..._Outlier_Bug.bdf.csv`).

## Our preference: keep units out of the column keys

More generally, we would prefer to keep units out of the column keys entirely
and instead rely on unit-aware serialization - e.g. the unit header line that
`pint-pandas` writes for CSV/Excel (a dedicated header row carrying each
column's unit alongside the plain, unit-independent key). That keeps the keys
stable and unambiguous while remaining fully unit-aware, and sidesteps the
`_second` vs `_s` (and similar) naming divergence above. A column would then be
e.g. `voltage` with unit `V` on the unit-header line, rather than `voltage_volt`.

## What would help

- A documented, normative key list (with canonical units), versioned so
  downstream tools can pin to it - ideally with units carried on a unit-header
  line rather than baked into the keys.
- Clean reference `.bdf.csv` (and/or parquet) files for the common quantities:
  test time, voltage, current, temperature, cycle/step index, and
  charge/discharge capacity + energy.

We are happy to contribute a mapping/adapter (and feedback) once the spec and
reference data stabilise.

g-kimbell:
Hi Simon, thanks for your interest in the project!

**Normative key list**
[...]

PR #66, opened 2026-07-14
docs(reference): canonical reference datasets with provenance manifest and scorecards

### LINKED ISSUE #52: Reference file non-compliance (step_index holds Step IDs; cumulative/net columns swapped) + add derived-column checks to validate()

### LINKED ISSUE #54: Reference files: Step ID placed in step_index column (FZJ, SINTEF Landt/Neware)

### LINKED ISSUE #55: Reference files: cumulative/net capacity & energy columns swapped (FZJ, BioLogic)

### LINKED ISSUE #56: Reference file: Neware G20M7 cycle_count is a constant 2π (+ minor cycle_count/monotonicity glitches)

jsimonclark:
## What  
Adds the canonical reference dataset collection: 13 BDF artifacts under `docs/examples/reference/`, each converted from a real raw vendor export with the current reader, plus the machine-readable provenance manifest (`datasets.json`), per-dataset scorecards rendered into a generated Reference Datasets docs page, the `scripts/build_scorecards.py` audit tool, and a CI test that re-validates every committed artifact on every run.

## Why  
#51 asked for stable, compliant reference data; #52/#54/#55/#56/#58 documented specific non-compliances in the previous reference files. This PR replaces the ad-hoc examples with a curated, provenance-tracked set: every artifact is paired to its raw source on Zenodo ([10.5281/zenodo.18986774](https://doi.org/10.5281/zenodo.18986774)) with URL, checksum, plugin id, and row/column inventory, and every artifact must pass `bdf.read(path, validate=True)` in CI. The scorecards make compliance human-reviewable: each dataset gets a verdict (PASS/WARN/FAIL) with plots, mapping tables, and an explicit "nothing is lost" unmapped-columns check.

## How  
Raw files are never committed. They live on the Zenodo record and are fetched through the content-addressed cache (repo policy, documented in the directory README). Artifacts are regenerated by `scripts/generate_zenodo_reference_bdf.py`; the exact invocation is recorded in `datasets.json`. Two artifacts keep real acquisition bugs on purpose (`__Time_Bug`, `__Outlier_Bug`) to exercise `bdf.clean`; they are format-compliant and marked in the manifest.

##Testing  
`tests/integration/test_reference_datasets.py` validates all 13 artifacts against the manifest in CI; the docs build renders the scorecard page from the committed PNGs and `datasets.json`.

## Notes   
Draft until two things land: (1) a follow-up commit adds a wall-clock cross-check to the scorecards that exposes the known Digatron time-unit defect (#65) as an honest FAIL on the FZJ dataset instead of a silent pass; (2) #63 (scale-aware derived-check tolerance) merges, which removes false-positive warnings on the FZJ file. Reference pairs for the new Arbin xlsx suite (Zenodo 0.8.0) are deliberately deferred until the accumulator remap after the next ontology release, to avoid regenerating them twice.

Closes #52. Closes #54. Closes #55. Closes #56. Refs #51, #58, #65.


jsimonclark:
Closing during a restructure: the canonical BDF artifacts will move to their own versioned Zenodo record (one version per bdf release) instead of being committed to git, with datasets.json as the manifest linking raw and artifact records. A fresh PR with that approach will replace this one and still resolve #52/#54/#55/#56. The branch stays; work continues there.

PR #67, opened 2026-07-15
Expand unit regex

g-kimbell:
Adds more characters to unit regex:
`·/*^%°℃ΩΩµμ⁰¹²³⁴⁵⁶⁷⁸⁹⁻ \-`

These are mostly natively handled by pint, so we don't need to do much more.

Now we capture units like
`A·h`, `A h`, `A*h`, `V/s`, `V s⁻¹`, `V / s`, `V s^-1`, `℃`, `µA`
which would get missed before

I also add the ohm symbol to unit alias just in case, pint only uses the Omega symbol, which is a different unicode character.

We also now capture a Bayestec temperature column in the existing dataset that was missed before.

### DIFF
--- a/src/bdf/spec.py
+++ b/src/bdf/spec.py
@@ -187,7 +187,7 @@ def get_unit_conversion(src_unit: str | None, dst_unit: str | None) -> tuple[flo
         scale = round(at_one - at_zero, 15)
         offset = round(at_zero, 15)
         return (scale, offset)
-    except pint.errors.PintError:
+    except (pint.errors.PintError, AssertionError):
         return None
 
 
--- a/src/bdf/spec.py
+++ b/src/bdf/spec.py
@@ -55,6 +55,7 @@ _UNIT_ALIAS = {
     "A.h": "Ah",
     "W.h": "Wh",
     "Ohm": "ohm",
+    "Ω": "ohm",  # 'ohm' character, 'Omega' character already understood by pint
 }
 
 # Bare "C" or "c" is ambiguous (Celsius vs Coulombs). This set lists the BDF
--- a/src/bdf/spec.py
+++ b/src/bdf/spec.py
@@ -167,8 +167,8 @@ def get_unit_conversion(src_unit: str | None, dst_unit: str | None) -> tuple[flo
     Returns:
         Tuple of (scale, offset) for conversion, or None if incompatible.
     """
-    src_bare = (src_unit or "").strip()
-    dst_bare = (dst_unit or "").strip()
+    src_bare = _normalize_unit((src_unit or "").strip())
+    dst_bare = _normalize_unit((dst_unit or "").strip())
     src_is_dim = src_bare in ("", "1")
     dst_is_dim = dst_bare in ("1", "")
     if src_is_dim or dst_is_dim:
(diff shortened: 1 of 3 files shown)


g-kimbell:
@jsimonclark @tomjholland ready for review (I still cant assign people)

jsimonclark:
Thanks @g-kimbell . I verified all the new spellings from your description parse correctly through pint (`A·h`, `A h`, `V s⁻¹`, `µA`, superscripts, the ohm-sign alias). 

But I also found a bug to fix before merge, and one new test request.

**Bug: a captured `℃` crashes the read.** The captured unit goes straight into `get_unit_conversion`, which only catches `pint.errors.PintError` But pint ≥ 0.25 raises a bare `AssertionError` on `℃`. You can reproduce with `get_unit_conversion("℃", "degC")`. Before this PR the regex couldn't capture `℃` so the path was unreachable; now a header like `T1[℃]` crashes `bdf.read` instead of converting. 

Suggest two small changes in `get_unit_conversion`:

- run `src_bare` and `dst_bare` through `_normalize_unit` (the `"℃": "degC"` alias already exists in `_UNIT_ALIAS`, it just isn't applied on this path), and
- change `except pint.errors.PintError` to `except (pint.errors.PintError, AssertionError)`, so anything pint chokes on degrades to "column unmapped" instead of crashing (same pattern `_pint_understands` already uses, for the same reason).

**Tests:** please add a small table-driven test covering the new spellings your description lists. Essentially that list as `(header_unit, expected_bdf_unit)` cases, including `℃`. That's what would have caught this in CI. I'll add a separate contract test for `get_unit_conversion` never raising on garbage input (that gap predates your PR).

Once the ℃ fix and the test are in, this is good to merge from my side.

g-kimbell:
@jsimonclark good catch, somehow with Python 3.14.3, pint 0.25.3 it works fine for me on main (11196af)
```
print(get_unit_conversion("℃", "degC"))
> (1.0, 0.0)
```
But I see the different errors on pint<0.25.
On pint<0.22 it raises an attribute error, but we have "pint>=0.22" in dependencies, so probably best to keep it restrictive.

I've added the new error catch and normalisation in `get_unit_conversion`, and added some more cases to `test_get_unit_conversion`

g-kimbell:
@jsimonclark in spec.py we already have
```
for _alias, _canonical in _UNIT_ALIAS.items():
    if _pint_understands(_alias, _canonical):
        continue
    # @alias adds another name to the existing unit (preserving offset
    # behavior for affine units like degC), unlike `name = degC`, which
    # would silently define a new non-offset unit.
    with contextlib.suppress(Exception):
        ureg.define(f"@alias {_canonical} = {_alias}")
```

Which runs on import and defines ℃ in pint, I think this is why I never had any error

How did you get the errors exactly?

The normalisation/extra catching may be unnecessary (though harmless)

jsimonclark:
Looks like its the python version, not pint. Same pint 0.25.3:

- Python 3.11.1: `ureg.Quantity(1, "℃")` → `AssertionError` (even after the `@alias` define, which succeeds but never gets consulted)
- Python 3.14.3: works

pint parses unit strings with python's `tokenize` module, which was rewritten in 3.12. The newer tokenizer tolerates `℃`, so the import-time alias resolves it; the older one asserts before the registry is consulted. That's why the alias loop helps on your machine but not on mine.

All checks green, changes look good. Merging.

PR #70, opened 2026-07-16
feat(io): reconcile elapsed-time unit against wall clock on read (#65)

pghege:
## What

Implements the time-reconciliation feature planned in #65: the FZJ HPPC Digatron reference file declares its test-time column in seconds but stores milliseconds. Because the timeline is self-consistent, nothing in the file alone reveals it — every single-column check passes. The defect only shows when the column's increments are compared with the independently recorded wall-clock (`Unix Time / s`) increments.

## Changes

**1. `src/bdf/_time_scale.py` (new)** — pure detection shared by read and validate: median ratio of elapsed-vs-wall-clock increments, matched against known unit factors (ns, µs, ms, min, h). Step-time resets, logging pauses, and NaNs drop out of the sample naturally (only rows where both clocks advance count); the median is robust to wall-clock 1 s-resolution jitter.

**2. `bdf.read`** — new `reconcile_time=True` parameter. When `Test Time / s` / `Step Time / s` increments disagree with the wall clock by a known factor, values are rescaled to seconds, the repair is recorded under `metadata["time_reconciliation"]` (column, ratio, detected unit, action), and a `UserWarning` is emitted. An unexplained ratio warns but leaves data untouched. Detection samples the first 100k rows so lazy reads stay cheap.

**3. `bdf.validate`** — reports the same mismatch as a warning-level `time_scale` finding in the derived-checks report, without modifying data — so already-converted artifacts (like the corrupted FZJ `.bdf` upload) fail loudly in reference-data gates instead of passing silently.

## Tests

15 new tests: detection math (ms/min factors, jitter, resets, pauses, NaNs, unexplained ratios, minimum samples), read-level repair + metadata for eager and lazy frames, opt-out, and validate-level reporting. Full unit suite: 591 passed.

Closes the detection/repair scope of #65; regenerating the FZJ artifact and greening its scorecard follows in the reference-datasets workflow.

Refs #65.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and 
 
 ## [Unreleased]
 ### Added
+- Time reconciliation on read (GH #65): `bdf.read` cross-checks `Test Time / s` and `Step Time / s` increments against wall-clock (`Unix Time / s`) increments; a column stored in the wrong unit under a seconds header (e.g. a Digatron export holding milliseconds) is rescaled to seconds, recorded under `metadata["time_reconciliation"]`, and announced with a `UserWarning`. Disable with `reconcile_time=False`. `bdf.validate` reports the same mismatch as a `time_scale` finding without modifying data.
 - repair.py ported to polars-native internals: fix_time and clean now accept polars (eager or lazy) or pandas frames and return the same kind, with identical numeric results across input types (pinned by the characterization suite). mypy re-enabled on the module; the CLI clean command no longer round-trips through pandas.
 - Arbin normalizers (csv, xlsx) map Charge/Discharge Capacity and Energy to the schedule_* terms from ontology 1.3.0, reflecting the operator-defined reset behavior Arbin confirmed; the columns previously landed on the never-resetting test-scoped terms.
 - CI pipeline with lint/type/tests/docs and build/twine checks.
--- a/docs/examples/metadata.ipynb
+++ b/docs/examples/metadata.ipynb
@@ -51,7 +51,13 @@
       "outputs": [],
       "source": [
         "# Read the raw source data and display the header\n",
-        "df, source_meta = bdf.read(\"https://zenodo.org/records/17295469/files/FZJ__INR21700__20250606__HPPC__25degC__Digatron.csv\")\n",
+        "# This export stores its time columns in milliseconds under seconds headers;\n",
+        "# bdf.read detects the mismatch against the recorded timestamps and raises\n",
+        "# unless we explicitly opt into the repair with reconcile_time=True.\n",
+        "df, source_meta = bdf.read(\n",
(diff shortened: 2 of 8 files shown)


pghege:
Converting to draft to sync with the pre-0.2.0 discussion — this is a working implementation of the reconciliation plan sketched in #65, but the design deserves group input before it moves forward. Two questions:

1. **Repair-on-read default**: as written, `read()` rescales a mismatched elapsed-time column automatically (with a `UserWarning` and a `metadata["time_reconciliation"]` record; `reconcile_time=False` opts out). The alternative is opt-in repair with detection-only by default. Given the general wariness about repair-based normalization, which do we want?

2. **Where the record lives**: the reconciliation record currently lands in the metadata dict returned by `read()`. Once the metadata unification work lands, should this become a field on the BDF metadata object instead?

The detection math (`_time_scale.py`) and the validate-side `time_scale` check are independent of both questions — happy to split those out if they're useful sooner, since they make the corrupted FZJ artifact fail loudly in the reference-data gates instead of passing silently.

g-kimbell:
Maybe we can do a simple check over all 3 time columns to catch any drift/discrepancy

`abs((unix_time - unix_time[0]) - (test_time - test_time[0]))` should be smaller than some tolerance factor

`abs((step_time - step_time[0]) - (test_time - test_time[0])) over step` should be smaller than some tolerance factor

> 1. Repair-on-read default

I would prefer loudly failing if this is detected. More useful in case a user is using that data for other stuff.

pghege:
Flipped as discussed: detection always runs, and a mismatch now raises `BDFValidationError` by default (warns when `validate=False`) with nothing modified. `reconcile_time=True` is the explicit repair flag: rescales known unit factors, records the repair in `metadata["time_reconciliation"]`, and warns. Unexplained ratios can't be repaired, so they stay loud even with the flag.

On the cumulative drift check `abs((unix_time - unix_time[0]) - (test_time - test_time[0]))`: I stayed with per-increment ratios (median over rows where both clocks advance) for two reasons. Cumulative drift accumulates false positives whenever the elapsed clock legitimately diverges from wall time — paused tests and logging gaps make the drift grow without any unit error, and several reference files have exactly that. And the ratio identifies *which* factor is wrong (ms vs min vs h), which the repair path needs. Your step_time-per-step formulation is effectively what the increment filtering does implicitly — resets produce non-positive increments and drop out of the sample. Happy to add a cumulative check on top if you see a defect class the ratio misses.

g-kimbell:
Sounds good to me

Drift checks could be a separate part of validation, there's a ton of checks we can add, but it can be done separately.

pghege:
Rebased onto the new read/scan API from #76: detection now lives in the shared _read, and reconcile_time is exposed on both read() and scan(). Same fsck behavior as discussed (raise by default, warn under validate=False, repair only with the explicit flag). Graham, when you have a moment this could use a fresh review since the rework.

pghege:
Going to merge this rather than hold it in the review queue: the rework implements exactly the design you voted for (detect and fail loudly by default, repair only with the explicit flag), and the release train is queued behind it. If anything looks off on a later read, happy to follow up in a patch, comments welcome any time.

PR #71, opened 2026-07-16
refactor: consolidate duplicated utilities; route column labels through the spec

pghege:
## What

First two steps of the edge-consistency plan (precursor to the `__init__.py` split; supersedes part of the motivation behind #11/#14). Behavior-neutral by construction — values are identical, call sites unchanged.

**Commit 1 — single owners for duplicated utilities.** Four copies of the http(s)-URL check (`__init__`, `registry_ld`, `file_utils`, `io`), three of the slug algorithm (`spec`, `validate`, `io._legacy_slugify`), and two of `ensure_dir` existed with no clear owner. Now: `spec._slugify` is the only slug implementation (it defines the slugs the synonym index is keyed by), `file_utils` owns `is_url` (already public there) and gains `ensure_dir`; everyone else imports.

**Commit 2 — spec-driven column labels.** `repair`, `validate`, `cli`, `visualize`, and `_explore` hardcoded `"Test Time / s"` / `"Voltage / V"` / `"Current / A"` — bypassing the ontology that owns those labels and contradicting the README's "documentation, code, and semantics cannot disagree" guarantee. They now derive from `COLUMN_ONTOLOGY.<quantity>.formatted_label`. A label rename in an ontology release now propagates everywhere from one place.

## Tests

Full unit suite green (575 passed), all pre-commit hooks pass. No new tests: no new behavior — the existing suite pins the label values transitively.

### DIFF
--- a/src/bdf/__init__.py
+++ b/src/bdf/__init__.py
@@ -108,11 +108,9 @@ if _enable_short_warnings():
 # small helpers
 # -------------------------------
 def _is_url(x: str) -> bool:
-    try:
-        u = urlparse(str(x))
-        return u.scheme in ("http", "https") and bool(u.netloc)
-    except Exception:
-        return False
+    from .file_utils import is_url
+
+    return is_url(str(x))
 
 
 def _resolve_source(
@@ -164,8 +162,9 @@ def _default_ingest_cache_dir() -> Path:
 
 
 def _ensure_dir(path: Path) -> Path:
-    path.mkdir(parents=True, exist_ok=True)
-    return path
+    from .file_utils import ensure_dir
+
+    return ensure_dir(path)
 
 
 def _parse_github_tree(url: str) -> tuple[str, str, str, str] | None:
--- a/src/bdf/__init__.py
+++ b/src/bdf/__init__.py
@@ -7,11 +7,11 @@ import warnings
 # mypy: ignore-errors
 from pathlib import Path
 from typing import Any, Literal, Optional
-from urllib.parse import urlparse
 
 import pandas as pd
 
 # light imports that never cause cycles
+from .file_utils import ensure_dir as _ensure_dir, is_url as _is_url
 from .io import read, save, scan  # spec-driven reader/serializer (the public read()/scan())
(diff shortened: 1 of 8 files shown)


src/bdf/__init__.py
```diff
@@ -108,11 +107,9 @@ def _enable_short_warnings() -> bool:
...
 # -------------------------------
 def _is_url(x: str) -> bool:
-    try:
-        u = urlparse(str(x))
-        return u.scheme in ("http", "https") and bool(u.netloc)
-    except Exception:
-        return False
+    from .file_utils import is_url
+
+    return is_url(str(x))
```
g-kimbell: I think you can do at the top
`from .file_utils import is_url as _is_url, ensure_dir as _ensure_dir`

src/bdf/__init__.py
```diff
@@ -164,8 +161,9 @@ def _default_ingest_cache_dir() -> Path:
 
 
 def _ensure_dir(path: Path) -> Path:
-    path.mkdir(parents=True, exist_ok=True)
-    return path
+    from .file_utils import ensure_dir
+
+    return ensure_dir(path)
```
g-kimbell: Same here
`from .file_utils import is_url as _is_url, ensure_dir as _ensure_dir`

src/bdf/_explore.py
```diff
@@ -7,6 +7,9 @@
 
 from . import spec
 
+X_DEFAULT = spec.COLUMN_ONTOLOGY.test_time_second.formatted_label
+Y_DEFAULT = spec.COLUMN_ONTOLOGY.voltage_volt.formatted_label
```
g-kimbell: Could do
`from .visualize import X_DEFAULT, Y_DEFAULT`
to have one source of truth

g-kimbell (commented):
LGTM! small suggestions

pghege:
All three applied, thanks. One note on the _explore one: it only became safe with #68 — visualize now lazy-imports matplotlib inside plot(), so importing the constants no longer drags matplotlib into _explore's import chain for non-[plot] installs. Rebased onto main while at it.

PR #69, opened 2026-07-16
docs(readme): fix placeholder citation, typos, duplicate quickstart block; complete FAQ

pghege:
## What

Front-page cleanup of README.md — no code changes.

- **Citation**: the Forrester study link pointed at a HubSpot PDF whose URL literally contained `DRAFT_PLACEHOLDER` (broken). Replaced with the study's official landing page: the 2024 Forrester Opportunity Snapshot commissioned by Monolith on AI for EV battery validation (same study: 165 automotive decision-makers, US/Europe).
- **FAQ**: the last entry ("How can I check if my file is a valid instance of BDF?") ended with no answer. Added one covering `bdf.validate()` and `bdf validate` on the CLI, including the ontology-defined derived-column checks from #53/#63.
- **Quickstart**: removed the duplicated interactive-exploration block (identical `bdf.explore` lines appeared twice).
- **Typos**: Pyton → Python, applicaiton → application.

Generated ontology term tables untouched (`scripts/generate_docs.py --check` passes).

PR #46, opened 2026-06-23
Remove legacy methods and update docs for spec-driven workflow

tomjholland:
## Summary
Finishes the migration to the spec-driven read/parse/normalize stack by deleting every legacy module it superseded (`detect.py`, `data_sources/`, `normalize/`, `units/`, `time.py`), routes unit handling exclusively through `spec.py`, and updates docs/examples to match: notebooks execute live via `myst-nb`, prose guides reflect the new `(frame, metadata)` read API, and a generated "Supported Plugins" reference page is added.

## Features
- `spec.load_version` now fetches and caches an ontology release when no local snapshot exists, instead of raising `ValueError` ([d1a5cea](https://github.com/tomjholland/battery-data-format/commit/d1a5cea323731ddd546142789b57275f2bf3a6be))
- Docs build executes example notebooks live via `myst-nb` instead of pre-converted `.rst` pages, with notebooks/fixtures/helpers moved under `docs/examples/` ([f258fd9](https://github.com/tomjholland/battery-data-format/commit/f258fd921509242ef10dafc0862c6b34b701c4ea))
- Auto-generated "Supported Plugins" reference page (`docs/plugins.rst`), built from `bdf.plugins.PLUGINS` by a new Sphinx extension on every docs build ([f3569b3](https://github.com/tomjholland/battery-data-format/commit/f3569b32630d5dce2f48e2b7e16b247debcdb298))
- CI/docs-deploy prefetch and content-address-cache notebook remote data sources before the docs build, since notebooks now execute live ([f1b69f4](https://github.com/tomjholland/battery-data-format/commit/f1b69f4cd79d5d53acda0f402e32583519031886))

## Refactoring
- Remove the legacy parsing/normalization/unit stack now that `bdf.read`/`detect`/`normalize` route through the spec-driven `io`/`plugins`/`table_normalizers` modules:
  - Inline `canonicalize_legacy_labels` (and its label sets) into `io.py`, its only consumer; `validate.py` now sources label sets from `spec` directly ([9888b19](https://github.com/tomjholland/battery-data-format/commit/9888b1912c792be57cf8f6639874c872613068ad))
  - Remove legacy `normalize` package; route read/detect/normalize to spec ([5843ac4](https://github.com/tomjholland/battery-data-format/commit/5843ac4da62e5c1eaecfd38906d349d13194b4b8))
[...]

PR #72, opened 2026-07-16
docs: pre-0.2.0 sweep — fix PyPI name in notebooks, complete API reference

pghege:
## What

Graham's 0.2.0 item 4 ("lots of docs still referring to the old methods — quick sweep before release"). Good news: a full audit of `docs/*.rst`, all 14 example notebooks, `CONTRIBUTING.md`, CLI help text, and source docstrings found that #46 already removed the legacy-method references. Only three confirmed-stale items remained:

- `docs/examples/visualization.ipynb`: `pip install bdf[hvplot]` → `pip install batterydf[hvplot]` (PyPI name is `batterydf`; import stays `bdf`)
- `docs/examples/registry.ipynb`: `pip install bdf` → `pip install batterydf`
- `docs/reference.rst`: `bdf.templates` is public API but was missing from the autosummary

Pairs with #69 (README fixes). Note: the reference page will need another touch when the read/scan API decision lands — deliberately not pre-empting that here.

## Verified

Everything else checked out: no references to removed modules (data_sources/normalize/units/detect/time), all documented API calls exist, install instructions in get_started.rst and README already correct.

### DIFF
--- a/docs/examples/registry.ipynb
+++ b/docs/examples/registry.ipynb
@@ -18,7 +18,7 @@
       "source": [
         "## Setup\n",
         "\n",
-        "Install BDF first: `pip install bdf`.\n"
+        "Install BDF first: `pip install batterydf`.\n"
       ]
     },
     {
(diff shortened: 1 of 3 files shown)


g-kimbell:
Ah that's great, I missed the `lazy=True` in all the examples so thought they were `to_pandas`ing lazyframes everywhere (mypy is still upset by it).

Will do another sweep before 0.2.0.

I'll also need to do another sweep after #68, we can merge in whatever order and sort it out.

PR #74, opened 2026-07-16
`read` warn when converting legacy columns

g-kimbell:
Adds `legacy` bool to `Syn`/`ResolvedColumn`

Gets set in `_build_bdf_normalizer`

Warns user when normalizing, e.g. with the `tests\data\bdf\legacy.bdf.parquet`:
```
UserWarning: Legacy BDF column labels detected and normalized to preferred labels: 'test_time_millisecond' -> 'Test Time / s'
```

Adds a small unit test for the bdf normalizer

This is a step towards removing `canonicalize_legacy_labels`, `load` etc. and just using `read` for all file types. We have a similar warning in `load` but not in `read`.

### DIFF
--- a/src/bdf/table_normalizers.py
+++ b/src/bdf/table_normalizers.py
@@ -56,6 +56,8 @@ class Syn(BaseModel):
     """True when no real-file sample exercises this synonym (see test_synonym_coverage)."""
     source_unit: str | None = None
     """Fixed source unit for exact, non-templated aliases."""
+    legacy: bool = False
+    """Raise a warning that this column is legacy and has been converted."""
 
     @model_validator(mode="before")
     @classmethod
@@ -130,6 +132,7 @@ class ResolvedColumn(BaseModel):
     datetime_fmts: tuple[str, ...] = Field(
         default=(), description="Datetime format strings for parsing timestamp columns."
     )
+    legacy: bool = False  # Resolved from a legacy column, warn user
 
     @classmethod
     def from_bdf_label(cls, bdf_label_key: str, src_header: str) -> tuple[str, ResolvedColumn]:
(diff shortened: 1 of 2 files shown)


jsimonclark:
@g-kimbell, found a false positive: a file with current preferred labels warns:

`'Test Time / s' -> 'Test Time / s'` 

because the deprecated `test_time_millisecond` label template (`"Test Time / {unit}"`) is the same string as the live term's and gets appended first with `legacy=True`, so it shadows the live synonym.

## Suggested fix 

In `_build_bdf_normalizer`: for deprecated terms, append the concrete label (`q.formatted_label`, e.g. `"Test Time / ms"`, with `source_unit=q.unit`) instead of the generic template. The deprecated terms have a fixed unit, so the template isn't needed, and the collision disappears while `"Test Time / ms"` still warns correctly. 

Please also add the regression test: modern prefLabel headers through `BDF_NORMALIZER` must emit no warning.

g-kimbell:
Good catch, sorry for missing the regression.

This was a little tricky - it failed because a label would check against:

1. `"Test Time / {unit}"` (deprecated)
2. `"Test Time / {unit}"` (current)

`"Test Time / s"` falsely matches generic deprecated first
 
Switching to the concrete label it checks:

1. `"Test Time / ms"` (deprecated)
2. `"Test Time / {unit}"` (current)

This works for both / s and / ms

But if the .ttl happened to be ordered differently, it would do

1. `"Test Time / {unit}"` (current)
2. `"Test Time / ms"` (deprecated)

Then `"Test Time / ms"` matches current and does not warn.

For now I've changed it to both use the concrete label **and** always append the deprecated terms to the table normalizer first.

We could consider trying to match against the concrete, preferred terms in a hashmap first, before checking explicitly deprecated and generic terms. It feels like iterating through deprecated terms first is a waste of time, but it is probably negligible. On top that, we should decide whether we warn users if they are not using the preferred BDF units. Right now there's no warning given if you hand over e.g. "Voltage / MV" and it converts the units. We should probably decide and deal with that in a separate issue/PR.

jsimonclark (approved):

PR #75, opened 2026-07-17
Add Codecov to CI

g-kimbell:
We should track our test coverage better

This PR adds Codecov to the CI. I suggest Codecov because I've had a nice experience with it on other projects - easy to set up, they say 'free forever' for public projects, you get a nice web UI, and you get a check in your PRs with useful info:
image

For now I've disabled comments in PRs (which I find very noisy), and do not enforce anything coverage for a PR to pass.

Later, we could enforce rules like 'changes must have 80% coverage and must increase the project coverage for a PR to pass' if we want.

@jsimonclark @pghege For this to work, I'd need an admin to install the codecov github app, allow it on this repo, and put CODECOV_TOKEN in the secrets. You can get the token from https://app.codecov.io.


jsimonclark:
@pghege , this one needs action from you

PR #78, opened 2026-07-17
Add json, ndjson, ipc/arrow/feather `TableParser`

g-kimbell:
In #76 I add the option to save to json/ndjson/ipc/arrow/feather, but we can't read it back.

This PR adds these as BDF table parsers, in the same way as the existing parquet parser.

Splitting into separate PR because this part is non-destructive and an easy merge.

`polars.read_json` only handles record-oriented json (which would make enormous files for us). I used `json.load` then put it in `LazyFrame(data)` which works for both record- and list- oriented json.

Also caught a subtle bug, table parsers were trying to match the last suffix (e.g. ".parquet") to all suffixes combined (e.g. ".bdf.parquet"). This means it would also fail if there are other dots in the filename. It was always silently falling back on the magic bytes for ".bdf.parquet" files. Now it checks all suffixes, last two suffixes, last suffix.

### DIFF
--- a/src/bdf/plugins.py
+++ b/src/bdf/plugins.py
@@ -33,7 +33,16 @@ from pydantic import BaseModel, ConfigDict, Field
 from .file_utils import is_url, read_head, resolve_source
 from .metadata_parsers import JsonSidecarParser, MetadataParser, MetadataSchema, TxtPreambleParser
 from .table_normalizers import BDF_NORMALIZER, NDA_NORMALIZER, NORMALIZERS, TableNormalizer
-from .table_parsers import DelimTxtParser, ExcelParser, MatParser, NDAParser, ParquetParser
+from .table_parsers import (
+    DelimTxtParser,
+    ExcelParser,
+    IPCParser,
+    JsonParser,
+    MatParser,
+    NDAParser,
+    NdjsonParser,
+    ParquetParser,
+)
 
 try:
     import yaml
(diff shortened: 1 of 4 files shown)


jsimonclark:
Thanks @g-kimbell , the suffix-matching fix is a nice catch and the parsers follow the existing patterns well. Two things before merge:

1. **JsonParser needs `encoding="utf-8"` in both `open()` calls.** Without it, Python uses the OS default encoding. On Windows that's cp1252, and JSON files with non-ASCII characters get silently corrupted rather than erroring: I tested it on a Windows machine and `°C` reads back as `Â°C`. JSON is UTF-8 by spec, so this is a one-word fix. CI never catches it because the runners are Linux.
2. Small one: the "Cannot be truly lazy" docstring got copy-pasted onto `NdjsonParser` and `IPCParser`, but `scan_ndjson` and `scan_ipc` are genuinely lazy, only the JSON parser isn't.

Good to go once the encoding fix is in.

g-kimbell:
Hi @jsimonclark , good catches.

I fixed the Ndjson parser lazy comment, IPC was okay.

I could recreate the `°C` -> `Â°C` behaviour when doing `ensure_ascii = False` in `json.dump`. I added a test that fails on in the previous state and is fixed with `encoding="utf-8"` in `JsonParser` open, plus a sanity check for the linux CI.

I also renamed `IPCParser` -> `IpcParser` to be consistent with the others.

jsimonclark (approved):

Issue #85, opened 2026-07-28
Wire a real .res sample into the test corpus and un-assume the arbin_res synonyms


pghege:
Follow-up to #60 and #84. The arbin_res normalizer's synonyms are marked assumed=True because no .res sample exists in the test corpus, so the synonym-coverage gate has nothing to check them against.

Abbta contributed a CC-BY 4.0 .res test file in the #60 discussion (constant-current test on a prussian white sodium-ion cell, https://github.com/user-attachments/files/30371584/Anwesa-20231222-PW-Na-ECSEC.zip) and Simon posted a second candidate. To close this:

1. Host the file somewhere SampleCase can fetch (Zenodo record alongside the other reference samples, or committed if small enough after extraction).
2. Add an arbin_res SampleCase with expected_columns covering the normalizer's headers.
3. Flip the assumed=True flags back off in ARBIN_RES.

Note the CI lesson recorded on #84: fork-PR pushes need maintainer CI approval; a stale green rollup let this land on main unchecked.

PR #86, opened 2026-07-28
Add minimal Abrin res test data

g-kimbell:
Adds a very small Arbin .res file for local tests, 8 rows, 17 kB gzipped.

I gzipped because the uncompressed version is >20x bigger. BDF can read the .res.gz directly without any modifications needed :)

Data is from me at Empa, fine to be released under this licence.


g-kimbell:
For now we just skip tests on Python 3.10.

I opened a PR in polars_access_mdbtools to lower the minimum Python version to 3.10, so maybe the skip won't be necessary in the future.

pghege (approved):
Exactly what #85 asked for, and the .res.gz working with no code changes is a nice validation of the compression path. Closes #85 when merged.

PR #87, opened 2026-07-28
Ndax data, small NdaParser changes

g-kimbell:
Adds a small (5 kB) .ndax file for local tests. From me at Empa, fine to release under this license.

Adds magic bytes for .ndax files, which are zips, so have b"PK" instead of b"NEWARE".

Changes `NDAParser` -> `NdaParser` to be consistent with other parsers `Mat` `Mpr` `Json` etc.


pghege (approved):
LGTM. Rename is fully covered (plugins, parser, tests, notebook all updated, nothing else references the old name) and PK magic follows the Excel parser precedent since ndax is a zip. Thanks for the Empa test file.

pghege:
Merged main into your branch to resolve a tail conflict in test_cases.py (#86 landed first, both PRs append a SampleCase). No changes to your code. Will merge on green.

PR #89, opened 2026-07-30
Update mdbtools and chores

g-kimbell:
Updates polars-access-mdbtools pin to 0.1.3, which now supports Python 3.10.

Removes the Python >=3.11 check for install and tests.

I also noticed the [arbin_res] group was not part of [all], so I added it and added a test that checks all non-dev deps are in [all], to avoid future drifts. I also added a check that ruff and mypy stay the same in pyproject and pre-commit, just in case.


pghege (approved):
LGTM. Nice catch on arbin_res missing from [all], and the drift-guard tests mean it stays caught.

Issue #91, opened 2026-08-03
save() metadata sidecar is never read back: round-trip silently loses metadata


pghege:
`save(df, path, metadata=...)` writes the metadata dict to a `<artifact>.metadata.json` sidecar, as documented. But `read()` on that artifact never picks the sidecar up, so the metadata survives on disk and silently vanishes from the API.

Reproduction on current main:

```python
import polars as pl
from bdf.io import read, save

df = pl.DataFrame({"Test Time / s": [0.0, 1.0], "Voltage / V": [3.7, 3.6], "Current / A": [0.1, 0.1]})
save(df, "x.bdf.csv", metadata={"cell_id": "X1"})
# x.bdf.csv.metadata.json now exists and contains {"cell_id": "X1"}

df2, meta = read("x.bdf.csv")
print(meta)  # {'source': 'bdf_csv'} - cell_id is gone
```

Two disconnects, for whoever picks this up:

1. Naming: `save()` writes `x.bdf.csv.metadata.json` (suffix appended), while `JsonSidecarParser` looks for `Path(path).with_suffix(".json")`, i.e. `x.bdf.json`. The two conventions never meet.
2. Wiring: the `bdf_*` plugins do not use a sidecar-aware metadata parser at all, so even a correctly named sidecar would not be read.

Why it matters for 0.2.0: the API freeze made `(df, metadata)` the read contract and kept the metadata dict for this release, with the richer metadata object (#47, #48) coming additively later. A write-only sidecar undermines that contract: anything a user stores through `save` is unrecoverable through `read`, with no warning.

Flagging for the rc window; describing the problem here first, solution discussion welcome below.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pghege:
Decision: not fixing this separately. Tom's metadata work (#48) reads metadata alongside the data as part of the typed metadata object, which closes this loop properly. 0.2.0 ships with the current behavior as a known limitation: the sidecar is written but not read back.

@tomjholland, two notes for when you pick this up:

1. The sidecar `save()` writes is `<name>.metadata.json` (full name appended, not `.with_suffix`), so the reader needs to match that.
2. `save(path)` without metadata leaves a previous save's sidecar in place. Once reading is wired up, a stale sidecar becomes wrong metadata attached to new data, so save should clear it in the same change.

g-kimbell:
[...]

PR #93, opened 2026-08-04
Use `with_suffix(".metadata.json")` for metadata

g-kimbell:
Saving metadata uses  `with_suffix(".metadata.json")`
`mydata.parquet` -> `mydata.metadata.json`
`mydata.bdf.parquet` -> `mydata.bdf.metadata.json`

We still cannot read the metadata back, but at least this removes the `mydata.parquet.metadata.json` currently.

Partially addresses #91

I think @tomjholland is still working on the metadata pydantic class / reading which we'd need to get proper round-trips working.


gabe-volta (approved):
LGTM. Also aligns JsonSidecarParser's expected name with what save() writes, and since no bundled plugin instantiates that parser, the change is behavior-safe. Good groundwork for the round-trip in #91.

One trade-off worth having on record: with_suffix means x.csv and x.parquet saved with the same stem share one x.metadata.json (the old appended names kept them distinct). Fine under the x.bdf.{format} naming convention where the .bdf. infix keeps stems unique per artifact, just noting it was a conscious choice. Merging after #94 to keep the module layout settled first.

PR #80, opened 2026-07-23
Update changelog for 0.2.0

g-kimbell:
Drafting as this will change before release.

Feel free to edit this branch directly.

This was mostly Claude scraping over git history, I've verified the parts I've touched. Could use some more eyes and trimming down.

PR #94, opened 2026-08-04
Lazy imports and rearrange modules

### LINKED ISSUE #90: Lazy imports: bdf/__init__.py eagerly imports pandas and the repair chain
`import bdf` (and therefore every CLI invocation) eagerly imports pandas, scipy (via repair.py), and the full io/validate stack. On a cold interpreter that is 10-20 s of import time before `bdf --version` prints anything, and the CLI never needs most of it for the invoked subcommand.

Proposal for 0.3.0:

- Move heavy imports (pandas, scipy, io, repair, validate) behind function-level or module `__getattr__` lazy imports in `bdf/__init__.py`.
- CLI subcommands import what they use at call time; `bdf --version` should touch nothing but the version string and ontology snapshot metadata.
- Since 0.2.0 removed pandas from the core paths, the eager top-level `import pandas` in `__init__.py` may be droppable outright.

No behavior change intended; public API surface (`bdf.read`, `bdf.scan`, `bdf.validate`, `bdf.clean`, ...) stays identical via lazy attribute resolution.

Observed while smoke-testing 0.2.0rc1 from TestPyPI on Python 3.14.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

g-kimbell:
This PR does a few things:

Makes `__init__.py` and `cli.py` imports lazy, closes #90.

I took `validate` and `ingest` out of `__init__` and put them in submodules.

⚠️ I renamed the modules `validate.py` ->`_validate.py` and `templates.py` -> `_templates.py`
Before, the module names conflicted with functions, e.g. we had `validate()` in `__init__`, which can cause unexpected behaviour.

I moved `BDFValidationError` to its own module, as there were lots of lazy imports for the error to avoid importing the validation functions. Now we can just import it at the top almost for free. Feels a bit silly to have by itself in a module for now, but maybe we'll have more warning/error types in the future.

Added tests for lazy import.
|                                       |Before (ms)|After (ms)|
|---------------------------|------------|-----------|
|`bdf --version`                | 2110 ± 40 | 230 ± 10 |
| `import bdf`                  | 2190 ± 80 | 130 ± 10 |
| `from bdf import read` |  2040 ± 50 | 660 ± 20 |


gabe-volta (approved):
LGTM. Measured on this branch: import bdf drops from ~2.9s to ~160ms with zero heavy modules loaded, and the CLI entry imports in ~130ms. Full unit suite passes. The module renames close out the name-conflict problem from the old monolith too.

One fix pushed to the branch: test_import_bdf_does_not_load_heavy_dependencies printed sys.argv[1:] but the subprocess got no arguments, so it always passed vacuously. Now it checks HEAVY_DEPS and still passes, so the guard is real.

Scope note: taking this into 0.2.0. It breaks the bdf.validate / bdf.templates module import paths, which is exactly the kind of break this release exists to make once. Changelog updated on #80 to match. The TestPyPI preview stays 0.2.0rc1; the tagged rc becomes rc2 so versions map to exactly one build.

g-kimbell:
Oops, thanks for catching the test issue!

Issue #48, opened 2026-06-23
Expand metadata element of field tuple to full Pydantic model


tomjholland:
Currently, the metadata we return from the `read()` method is pretty barebones. I think the natural extension of this is for it to return a complete pydantic model of the metadata schema (currently built as data classes in `metadata.py`). The metadata is then read by the `MetadataParser` class and is then returned pre-filled in the pydantic model, to which the user can make additions and then save as jsonld. In the case where the user does not have their own bdf-compliant metadata, the already built (for start time extraction) parsers glean as much information as is available, by searching the file header or mapping the user's own sidecar json file.

Issue #92, opened 2026-08-03
No convention for linking a BDF file to its cell and test


pghege:
A BDF file carries no reference to the cell or test it belongs to. Today the association exists only through the Datastore directory layout (batteries/<cell_id>/...), which is one realization with no stated contract behind it. Copy a file out of that layout and nothing in or near the file says which cell produced it, what test it was part of, or where its metadata lives.

This is the flip side of BDF being lightweight. Container formats (HDF5, DICOM) enforce these associations physically by putting everything in one file. BDF chose loose files that any tool can open, which means associations have to be carried by convention instead. That works, but only if the convention is written down: the spec states the intent (a BDF file declares which cell and test it belongs to, in a canonical place), and then any realization that honors it interoperates, whether that's a field in the metadata object, a sidecar key, or an embedded entry.

What's missing is that written intent. Concretely:

- No identifier fields are defined (cell id, test id, or equivalent).
- No canonical place to look for them is defined.
- validate() has no notion of them, so nothing can check or enforce the convention.

Related: the metadata schema work in #47 / #48 is the natural home for the identifier fields themselves, and #91 covers the read-back mechanics. This issue is about defining the linkage convention, not the storage.

Describing the problem here first; solution discussion welcome below.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

g-kimbell:
As discussed on 2026-08-04, best to put cell info and test info as fields in the metadata object (in addition to 'raw' metdata from original file).

PR #96, opened 2026-08-17
Battinfo metadata and normalization rework

tomjholland:
This PR formalises the structure of metadata returned on a `bdf.read()`/`bdf.scan`, closing #48.

(Do not fear, most of the 10,000 lines are json schema files 😅)

The returned object is a pydantic model with the following structure:

```python
class Metadata(_RecordModel):
    battinfo_test: BattinfoTest = Field(default_factory=BattinfoTest)
    battinfo_cell: BattinfoCellInstance = Field(default_factory=BattinfoCellInstance)
    battinfo_channel: BattinfoChannelInstance = Field(default_factory=BattinfoChannelInstance)
    battinfo_equipment: BattinfoEquipmentInstance = Field(default_factory=BattinfoEquipmentInstance)
    battinfo_test_protocol: BattinfoTestProtocol = Field(default_factory=BattinfoTestProtocol)
    bdf: BdfReadInfo = Field(default_factory=BdfReadInfo)
    extras: dict[str, Any] | None = None
    raw: dict[str, Any] | str | None = None
```

### Most metadata mirrors BattINFO exactly

- no need for two standards
- most will stay empty on a normal read, in `save()` the json is dumped with `exclude_defaults=True` so unset fields won't clutter the output
- the user then has a complete BattINFO-compliant metadata schema that they can fill
- Pydantic models validate the types specified in the BattINFO schema, preventing an invalid item from later unexpectedly causing BattINFO incompatibility. BattINFO's tags haven't been updated in a while so this is currently a bit of a hack, checking that the content of the latest git commit on main is identical to the existing content. There is definitely a better way to do this once there is a versioned pin at a w3id address; currently I think the schemas are un-versioned, so cannot be used as a static ref
  - means there are a few minor niggles to iron out e.g. unix times are constrained to integers, losing ms accuracy if it exists in the source file

### BattINFO is not a project dependency (since the python package already has BDF as a dependency of it)

- BattINFO schemas are packaged with BDF and updated with `scripts/update_battinfo.py`, similarly to the BDF ontology, verified as the latest version with a test.
- Pydantic models are auto-generated from schema files using [datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator) (MIT licenced), so no manual intervention needed when BattINFO updates.

### BDF deals only with sidecar `.json` files, never `.jsonld`

I think this is the safest option, making BDF the "simple default". Those looking to publish with linked metadata should do so with the BattINFO package so that iris can be consistently minted etc. I will update the method of that package to accept this new BDF metadata object (which by design should be trivial) once we are happy here -> I think this means deprecating/updating the prior metadata handling methods, so that we do not have multiple competing paths.

### Unified normalisation

Mechanics for normalisation of numerical and date columns and metadata fields is unified into `normalisation.py`. This mostly impacts datetimes, which now have separate handlers for absolute time (epoch), relative time (epoch but to be interpreted only from the start of the experiment) and elapsed time (HH:MM:SS only):

previously:

```python
NEWARE = TableNormalizer(
    test_time_second=(
        DateTimeSyn(syn=Syn(hdr="Total Time", assumed=True), fmts=_NEWARE_DT_FMTS),
        Syn(hdr="Total Time({unit})"),
        ...
    ),
    voltage_volt=(
        Syn(hdr="Voltage({unit})"),
        ...
    ),
    current_ampere=(
        Syn(hdr="Current({unit})"),
        ...
    ),
    unix_time_second=(
        DateTimeSyn(syn=Syn(hdr="Date"), fmts=_NEWARE_DT_FMTS),
        DateTimeSyn(syn=Syn(hdr="DateTime", assumed=True), fmts=_NEWARE_DT_FMTS),
        DateTimeSyn(syn=Syn(hdr="Date_Time", assumed=True), fmts=_NEWARE_DT_FMTS),
    ),
```

now:

```python
NEWARE = TableNormalizer(
    test_time_second=(
        Syn(hdr="Total Time", assumed=True, normalization=RelativeTimeNormalization(formats=_NEWARE_DT_FMTS)),
        Syn(hdr="Total Time({unit})"),
        ...
    ),
    voltage_volt=(
        Syn(hdr="Voltage({unit})"),
        ...
    ),
    current_ampere=(
        Syn(hdr="Current({unit})"),
        ...
    ),
    unix_time_second=(
        Syn(hdr="Date", normalization=AbsoluteTimeNormalization(formats=_NEWARE_DT_FMTS)),
        Syn(hdr="DateTime", assumed=True, normalization=AbsoluteTimeNormalization(formats=_NEWARE_DT_FMTS)),
        Syn(hdr="Date_Time", assumed=True, normalization=AbsoluteTimeNormalization(formats=_NEWARE_DT_FMTS)),
    ),
```

and the same structure can then be used for metadata, where we declare a rule as a target, an extraction, and a normalization:

```python
BIOLOGIC_MPT = Plugin(
    table_parser=DelimTxtParser(...),
    metadata_parser=TxtPreambleParser(
        magic=("bt-lab ascii file", "ec-lab ascii file"),
        rules={
            METADATA.battinfo_test.test.started_at: RegexRule(
                pattern=re.compile(r"Acquisition started on\s*:\s*(.+)"),
                normalization=AbsoluteTimeNormalization(formats=_BIOLOGIC_DT_FMTS),
            ),
        },
    ),
)
```

Supporting cleanups: validates a timezone without building a frame, renders the plugin metadata rules straight from the declarations, and rebuilds the example notebooks against the new API.

### Month name translation

New [babel](https://github.com/python-babel/babel) (BSD-3 licence) dependency, which intercepts all written month names from dates, translates them and replaces them with the month number for parsing.

### User-control of their day-month order preference

Formats like `%d/%m/%y` can be indistinguishable from `%m/%d/%y`, and cyclers may change their convention depending on the location/language of the user. Added a new argument to `bdf.read()` etc., `day_month_order: Literal["day_first", "month_first"] | None = None`

- if None: formats are parsed in exactly the order they are defined in the plugin (first wins)
- if `day_first`/`month_first` the order of the month or day is swapped to match the convention (no-op if the format is already defined this way)

# Remaining
- Make corresponding PR on BattINFO repo + full integrated example
- Deprecate old metadata handlers


pghege (commented):
Thanks for posting the PR. The structure is strong: the sidecar round-trip closes #91, the generated models plus the upstream-drift network test solve the schema sync problem properly, day_month_order defaulting to None keeps every plugin's declared behavior untouched, and the parity tests pinning the normalization rework are the right insurance for a change this size. The changelog covers the breaking change too. Nice work.

One find that needs a fix before merge:

save() without metadata leaves a previous save's sidecar in place, and read() now restores it. Reproduced on this branch:

```python
m = Metadata()
m.battinfo_test.test.name = "CELL-A-run1"
save(df1, "x.bdf.csv", metadata=m)
save(df2, "x.bdf.csv")          # different data, no metadata
_, meta = read("x.bdf.csv")
meta.battinfo_test.test.name    # "CELL-A-run1", attached to df2's data
```

Nothing warns, so the old cell's identity rides along with the new cell's data. Suggest save() without metadata should refuse when an existing sidecar sits at the target: raise naming the file and the two explicit outs, pass metadata= from your read to keep or update it, or metadata=Metadata() to clear it. Silent retention attaches old identity to new data, and any deleting default destroys metadata the user may have meant to keep. An error at save time loses nothing and matches the raise-by-default posture read() took for time reconciliation. The explicit-empty path composes with your exclude_defaults handling: after it, no sidecar exists and read falls through to the plugin parser, so a sidecar exists exactly when the artifact has non-default metadata. Longer term the sidecar could carry a binding to its data file so read can verify the pairing, which is #92's linkage question showing up again, but refuse-by-default is enough for this PR. Open to a different mechanism if you have one in mind.

One sequencing question for the group:

The Metadata model is exactly what we agreed in #76, the wrinkle is how it lands. #76 framed it as coming additively without breaking the signature, and as built it replaces the dict, so anything reading metadata["source"] breaks. Three ways to land it: fold into 0.2.0 and break once (rc1 was only the TestPyPI preview, rc2 would carry this), add a deprecated dict-style shim for the old keys, or hold for 0.3.0. I lean folding into 0.2.0, it is the breaking release and shipping the dict only to break it one release later defeats the point. Simon, want your read before we sequence the tag.

tomjholland:
> Thanks for posting the PR. The structure is strong: the sidecar round-trip closes #91, the generated models plus the upstream-drift network test solve the schema sync problem properly, day_month_order defaulting to None keeps every plugin's declared behavior untouched, and the parity tests pinning the normalization rework are the right insurance for a change this size. The changelog covers the breaking change too. Nice work.
> 
> One find that needs a fix before merge:
> 
> save() without metadata leaves a previous save's sidecar in place, and read() now restores it. Reproduced on this branch:
> 
> ```python
> m = Metadata()
> m.battinfo_test.test.name = "CELL-A-run1"
> save(df1, "x.bdf.csv", metadata=m)
> save(df2, "x.bdf.csv")          # different data, no metadata
> _, meta = read("x.bdf.csv")
> meta.battinfo_test.test.name    # "CELL-A-run1", attached to df2's data
> ```
> 
> Nothing warns, so the old cell's identity rides along with the new cell's data. Suggest save() without metadata should refuse when an existing sidecar sits at the target: raise naming the file and the two explicit outs, pass metadata= from your read to keep or update it, or metadata=Metadata() to clear it. Silent retention attaches old identity to new data, and any deleting default destroys metadata the user may have meant to keep. An error at save time loses nothing and matches the raise-by-default posture read() took for time reconciliation. The explicit-empty path composes with your exclude_defaults handling: after it, no sidecar exists and read falls through to the plugin parser, so a sidecar exists exactly when the artifact has non-default metadata. Longer term the sidecar could carry a binding to its data file so read can verify the pairing, which is #92's linkage question showing up again, but refuse-by-default is enough for this PR. Open to a different mechanism if you have one in mind.
> 
> One sequencing question for the group:
> 
> The Metadata model is exactly what we agreed in #76, the wrinkle is how it lands. #76 framed it as coming additively without breaking the signature, and as built it replaces the dict, so anything reading metadata["source"] breaks. Three ways to land it: fold into 0.2.0 and break once (rc1 was only the TestPyPI preview, rc2 would carry this), add a deprecated dict-style shim for the old keys, or hold for 0.3.0. I lean folding into 0.2.0, it is the breaking release and shipping the dict only to break it one release later defeats the point. Simon, want your read before we sequence the tag.

Thanks Gabe, good catch!

Latest updates should resolve this. The four cases:

- metadata=None and a sidecar exists: raise, before the table is
  written, so neither file changes.
- metadata=None and no sidecar: write the table and create no sidecar,
  as before.
- a Metadata that states nothing: delete the sidecar, or write none
  where none exists. That documented clear path never cleared anything
  before. It wrote no file and left the old one in place.
- a Metadata that states something: write the sidecar, replacing any
  existing one, as before.

We should now also be protected against empty `Metadata()` objects (that would delete an existing sidecar) being unexpectedly produced internally:
- If a .json exists but fails to read for any reason, we raise to stop the read - previously an empty `Metadata()` would be returned

pghege (approved):
Verified the fix on the branch, all four save cases behave as described: the refuse fires before the table is written so nothing changes on the error path, explicit empty clears the sidecar, round-trip returns what was saved, and a malformed sidecar raises instead of masquerading as empty. The unreadable-sidecar raise is a good call, it closes the accidental-clear path too.

On sequencing: folding this into 0.2.0. It is the breaking release, and shipping the dict just to break it one release later defeats the point. Your changelog entries are already in the 0.2.0 section so nothing moves. The tag becomes 0.2.0rc2 and goes out once the reference bundle lands. Simon, if you see a problem with that, flag it before the tag.

One nit, not blocking: test_committed_package_matches_a_fresh_render fails with "No module named datamodel_code_generator" on a machine without the dev tooling installed, an importorskip would make it skip cleanly instead.

Approving.

PR #101, opened 2026-08-23
Minor refactor of save mechanism enabling sink_* methods

tomjholland:
Polar's `sink_*` methods allow writing data to the disk without it first being completely collected into RAM. This PR changes the backend structure of the `save()` method to a dict keyed by format rather than an if-statement chain.

### DIFF
--- a/src/bdf/io.py
+++ b/src/bdf/io.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import json
 import warnings
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal, cast
+from typing import TYPE_CHECKING, Any, Literal, NamedTuple, cast
 
 if TYPE_CHECKING:
     import pandas as pd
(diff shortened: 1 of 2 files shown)


pghege (commented):
Verdict: request changes for one required fix; the refactor itself is sound.

**The mypy CI failure** is the only blocker: `io.py:499` — `getattr(frame, spec.sink if isinstance(frame, pl.LazyFrame) else spec.write)` is guarded by the preceding collect-if-no-sink, so `spec.sink` can't be None when the lazy branch is taken, but mypy can't see the cross-statement invariant. An explicit lazy branch with `assert spec.sink is not None` (or a two-branch rewrite) fixes it. The offline/network jobs show "skipping" only because pre-commit gates them.

Verified good:
- Ran `tests/unit/test_io.py` on the branch with all extras: 72 passed, including the new lazy roundtrips across 6 formats × 5 codecs; the `polars>=1.41.2` floor matches what the compressed lazy sinks need.
- The #96 refuse-by-default sidecar guard still leads `save()` before format detection or any file creation (verified on the branch, `io.py:477-485`); `_write_sidecar` is a pure extraction.
- Unknown-format still raises before touching disk; the deleted `else` branch was dead code.
- Two quiet improvements worth keeping: compressed-xlsx now raises before creating an empty `.gz`, and `mkdir(parents=True)` moved after the checks, so invalid saves no longer create parent dirs.

One behavior note, not blocking: on main a failing lazy computation raised before the target file existed; with `sink_*` the plan executes after `open_compressed` created the target, so a failing lazy plan now leaves a partial artifact behind (`io.py:501-507`, no cleanup-on-failure in `save()`). #102 establishes the write-to-temp-then-rename pattern that would close this — worth a follow-up rather than blocking the streaming win.

jsimonclark:
@tomjholland , do you have time to address these changes? I can support, if needed.

tomjholland:
> @tomjholland , do you have time to address these changes? I can support, if needed.

Yep I can sort this morning

jsimonclark:
Approving for merge. @tomjholland 's fix resolves the mypy blocker from @pghege 's review, behavior is unchanged in all three write paths, and the full CI matrix is green.

I'll open an issue (#104 ) for the partial-artifact-on-failed-sink note, and the changelog line for streaming saves can ride #100.

pghege (commented):
Thanks Tom, the registry shape is a real improvement and the sidecar semantics came through intact. One catch before merge: the streaming benefit doesn't materialize yet. validate_df runs unconditionally in save and its first step is _to_polars_lazy(df).collect(), so a lazy save materializes the whole frame for validation and then sink_* re-executes the query from scratch, full RAM cost plus double compute. Suggest a single-compute rule: a lazy frame with validate=True collects once, validates, and writes eagerly; sink_* engages only when validate=False, documented as skipping validation and its warnings. That keeps the no-RAM path real for the bulk-pipeline case it's meant for. Happy to merge with that change, or talk me out of it.

pghege:
On reflection, this can wait, opened https://github.com/battery-data-alliance/battery-data-format/issues/108 to track it rather than holding the PR, behavior is correct and the cost question is a follow-up. Simon's approval stands, merging.

PR #102, opened 2026-08-23
Fix/download temp file cross filesystem

tomjholland:
## Summary

`bdf.fetch` downloaded a file into the system temporary directory, then renamed
it onto the cache path. A rename is atomic only inside one filesystem, so every
environment that splits `/tmp` from the cache directory raised
`OSError: [Errno 18] Invalid cross-device link` on any cache miss. The download
now writes its temporary file beside the destination, which restores the atomic
rename and makes the cache work under a tmpfs `/tmp`, a separate `/home`
partition, an NFS home directory, or a container bind mount.

## Fixes

- `_download` creates its temporary file in the destination directory instead of
  the system temporary directory, so the rename onto the cache path stays inside
  one filesystem and never raises `EXDEV`. This matches the pattern already used
  in `spec.py`, `file_utils.py`, and `scripts/update_battinfo.py`
  ([447572a](https://github.com/tomjholland/battery-data-format/commit/447572ad50284c1e0edd587c88749423205c7f9a))
- A failed download removes its temporary file rather than leaving a `.tmp`
  remnant in the cache directory, which matters because `fetch_url` retries each
  URL and falls through to `alt_urls`
  ([447572a](https://github.com/tomjholland/battery-data-format/commit/447572ad50284c1e0edd587c88749423205c7f9a))
- A large download no longer passes through the system temporary directory, so a
  RAM-backed tmpfs `/tmp` cannot fail the fetch with `ENOSPC`
  ([447572a](https://github.com/tomjholland/battery-data-format/commit/447572ad50284c1e0edd587c88749423205c7f9a))

## Tests

- `TestCacheDir::test_fallback_when_unset` and `test_fallback_when_empty` used a
  hard-coded `/tmp/bdf-fallback-xyz` path and failed wherever `/tmp` is not
  writable. Both use the `tmp_path` fixture now, and both assert that
  `cache_dir` creates the directory
  ([8213009](https://github.com/tomjholland/battery-data-format/commit/82130094954eef32946924666644f619f0d68f69))

## Verification

`uv run pytest -m "not network"` gives 853 passed, 19 skipped, 17 deselected.
The deselected cases fetch from Zenodo and need network access.

### DIFF
--- a/src/bdf/fetch.py
+++ b/src/bdf/fetch.py
@@ -79,14 +79,31 @@ def _safe_cache_name(url: str, filename: Optional[str]) -> str:
 
 
 def _download(url: str, dest: Path, timeout: int = 120) -> None:
-    with requests.get(url, stream=True, timeout=timeout) as r:
-        r.raise_for_status()
-        with tempfile.NamedTemporaryFile(delete=False) as tmp:
-            for chunk in r.iter_content(chunk_size=1 << 20):
-                if chunk:
-                    tmp.write(chunk)
-            tmp_path = Path(tmp.name)
-    tmp_path.replace(dest)
+    """Download ``url`` to ``dest`` through a temporary file in the same directory.
+
+    A rename is atomic only inside one filesystem. The system temporary
+    directory is often on another filesystem, so the temporary file must sit
+    beside ``dest``.
+
+    Args:
+        url: Source URL.
+        dest: Destination path.
+        timeout: Per-request timeout in seconds.
+    """
+    dest.parent.mkdir(parents=True, exist_ok=True)
+    with tempfile.NamedTemporaryFile(dir=dest.parent, delete=False, suffix=".tmp") as tmp:
+        tmp_path = Path(tmp.name)
+    try:
+        with requests.get(url, stream=True, timeout=timeout) as r:
+            r.raise_for_status()
+            with open(tmp_path, "wb") as fh:
+                for chunk in r.iter_content(chunk_size=1 << 20):
+                    if chunk:
+                        fh.write(chunk)
+        tmp_path.replace(dest)
+    except BaseException:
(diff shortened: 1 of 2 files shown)


pghege (commented):
Verdict: approve. The red CI is not this PR.

- Fix verified: `NamedTemporaryFile(dir=dest.parent)` makes `tmp_path.replace(dest)` a same-directory atomic rename — EXDEV impossible; matches the existing pattern in `file_utils.py:132`. `except BaseException: unlink` covers mid-stream failures and interrupts, which matters because `fetch_url` retries across `alt_urls` and previously leaked one orphan per failed attempt. Concurrent fetches each get unique temp names and last-writer-wins atomically; sha256 still verifies on `dest` post-rename. All 12 fetch tests pass on the branch.
- CI diagnosis: one root cause, unrelated to the diff — `test_bundle_matches_the_upstream_default_branch` hit a GitHub 403 rate limit (the exact flake #103 addresses), and the 3.10/3.12 jobs finished 285 passed / 0 failed before fail-fast cancelled them. The PR's own paths are green everywhere.
- Recommend folding this into 0.2.0rc2: it fixes a hard crash (`OSError: EXDEV`) on every cache-miss fetch wherever `/tmp` and the cache sit on different filesystems — tmpfs `/tmp` is the default on much of modern Linux, plus containers with bind-mounted caches. Small, test-covered, no API change; shipping an rc with a known crash whose fix exists undercuts the rc. Practical order: land #103, rebase this so the flake can't re-red it, then include in the rc2 tag.

Nit, non-blocking: a SIGKILL leaves a stray `tmp*.tmp` in the cache dir forever (the old /tmp location at least got OS-cleaned); stray files are inert to exact-name lookups, but a startup sweep would close it.

PR #103, opened 2026-08-23
Fix/skip tests without deps

tomjholland:
Add skip markers on tests that require optional dependencies. Includes moving the pybamm test dependency into a separate `test-optional` group. 

Reduces likelihood of rate-limits being hit by tests guarding against stale BattINFO and bdf ontologies by creating a new test group "freshness" that runs before all other tests. These tests are then excluded from the parallel matrix runs across python versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,9 +68,9 @@ jobs:
       - name: Install MDB Tools
         run: sudo apt-get install -y mdbtools
       - name: Install dependencies
-        run: uv sync --all-extras
+        run: uv sync --all-extras --group test-optional
       - name: Run tests (offline)
-        run: uv run --all-extras pytest -q -m "not network" --cov=bdf --cov-report=xml
+        run: uv run --all-extras --group test-optional pytest -q -m "not network" --cov=bdf --cov-report=xml
       - name: Upload coverage
         uses: codecov/codecov-action@v5
         with:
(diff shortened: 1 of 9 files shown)


pghege (commented):
Verdict: approve, with one follow-up request in this PR's own territory.

Verified: the marker algebra is airtight — all four freshness tests carry `network + live_network + freshness`; offline runs `not network`, the network job runs `network and not freshness`, the freshness job runs `-m freshness` once; no marker combination falls into a no-job gap. pybamm still runs across the full 3.10–3.14 offline matrix via `--group test-optional`, and the network job dropping pybamm is safe (top-level importorskip, no network marker on that file). CI green including the new freshness job.

The follow-up: the yadg skips use `pytest.mark.skipif(pytest.importorskip("yadg", ...) is None, ...)` at module scope in `tests/integration/test_cases.py` — `importorskip` raises module-level Skipped at call time, so without yadg the whole module vanishes at collection (reproduced on the branch: `--collect-only` yields no tests, and arbin/neware/novonix/maldives cases all disappear behind one "fastnda not installed" line). CI is unaffected because every job installs `--all-extras`, and the pattern pre-exists (fastnda/fastexcel), but a PR titled "skip tests without deps" is the right place to fix the mechanism: `skipif(importlib.util.find_spec("yadg") is None, ...)` gives true per-case skips — the new MDB marker in the same file already does it right.

Two notes for the team, no objection: the freshness gate means an upstream ontology change now turns all CI red with zero downstream results (deliberate, documented — but a behavior change worth knowing), and the freshness job still calls the GitHub API unauthenticated, so the 403 flake class is rarer, not gone; plumbing `GITHUB_TOKEN` through would end it — candidate follow-up issue.

PR #107, opened 2026-08-28
Refresh BattINFO bundle to upstream b9b49bb

pghege:
Ran scripts/update_battinfo.py against upstream main, the freshness test caught the drift after this week's upstream changes.

Mostly mechanical (regenerated schemas + models + VERSION), with one judgment call for your eyes @tomjholland: upstream's Quantity schema now requires a unit alongside any value, so a bare number in a Quantity-typed conditions slot no longer validates. The conditions test now models that (value + unit_text object). If you'd rather coerce or handle it differently in the parsers, say so and I'll adjust.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pghege:
Merging now to unblock the rc tag, the strictness here is upstream's own contract (Simon's schema change), so adapt-and-enforce is the default-correct call. @tomjholland if you'd rather the parsers coerce bare numbers, happy to take that as a follow-up.

Issue #105, opened 2026-08-25
Review selected BattINFO schema


tomjholland:
The selected BattINFO schema to load into BDF were chosen as the minimum (8 of the 38) set to reduce bloat, so not everything is included. Before `v0.2.0` I thought I would just raise this issue so that @jsimonclark you can advise on obvious missing schema. 

`MANAGED_SCHEMA_PATHS` in `scripts/update_battinfo.py` contains the list.

## What is bundled

- five entity schemas, one per record on `bdf.metadata.Metadata`: `test`,
  `cell-instance`, `channel`, `equipment`, `test-protocol`
- `cell-canonical`, `modules/common/quantity`, and
  `modules/common/quantitative-properties`, which those five refer to

## The 30 excluded files

- **Entity records**: `battinfo.base`, `dataset`, `organization`,
  `parameter-set`, `ingest-manifest`
- **Spec halves**: `cell-spec`, `equipment-spec`, `electrode-spec`,
  `electrolyte-spec`, `separator-spec`, `housing-spec`,
  `current-collector-spec`, `material-spec`
- **Instance halves**: `electrode`, `electrolyte`, `separator`, `housing`,
  `current-collector`, `material`
- **Modules**: `common/provenance`, and the ten under `components/`
  (`specification`, `electrode`, `electrode-coating`, `electrolyte`,
  `separator`, `current-collector`, `material-component`, `material-mixture`,
  `cell-instance-reference`)
- **Profile**: `profiles/batterypass`

Upstream also publishes non-schema assets that BDF excludes: the curated
property, unit, and material maps under `assets/mappings/`, plus
`assets/shapes/` and `assets/vocab/`.

## My thoughts

We should probably add the `*-spec` halves of what we already have. I intentionally excluded `dataset` under the principle that BattINFO itself should be handling published works, but I'm happy to include if we think it's valuable for round-tripping published data.

## What an addition changes

An addition touches `MANAGED_SCHEMA_PATHS`, the bundled files, the generated package, and `tests/unit/test_battinfo_contract.py`. A new record field on `bdf.metadata.Metadata` also changes the sidecar shape. As an addition, this would probably not be breaking, but worth getting right before we release.

tomjholland:
#106 is tied into this - currently a schema change would be breaking, which we can guard against by pinning the schema version in the metadata

jsimonclark:
Thanks for setting this up @tomjholland. 

[...]

Issue #98, opened 2026-08-22
Time-scale check flags minute-resolution wall clocks as ~0.5x mismatch


pghege:
The derived-consistency check compares elapsed-time increments against wall-clock increments row by row. When the source's datetime column has minute resolution, consecutive rows share a timestamp and then jump 60 s, so the increments read {0, 60} while test_time advances in true 10-30 s steps. The median ratio lands near 0.5x with 'no known unit matches this ratio'.

Concrete case: the regenerated maccor reference file (faraday__lg-INR21700M50-2019-002__2019-06-02__rate__25degC__maccor.bdf.csv) from the canonical bundle. unix_time_second diffs: 3420 zeros, 2889 sixties; test_time diffs: clean 10 s / 30 s steps. The data is faithful to its source; the wall clock is just quantized.

Proposal: detect wall-clock quantization (increments concentrated on {0, k}) and compare over windows of >= k rather than per-row, or annotate the finding as resolution-limited instead of unit-mismatch. Post-0.2.0.

The file enters KNOWN_NONCOMPLIANT referencing this issue until the checker understands resolution.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #115, opened 2026-09-01
battinfo_dataset record, sidecar version stamps, manifest integrity

### LINKED ISSUE #106: Pin BDF and BattINFO version numbers to metadata
Currently we don't have a field that pins the versions of BDF or BattINFO used to write a file. The structure could be:

- BDF package version
- BDF ontology version (since an updated package *should* be able to avoid breaking changes in the schema by live-updating the version of the ontology that it loads)
- BattINFO schema version (similarly, for any version of the bdf python package we should be able to load any schema. This should guard against a schema change becoming a breaking change for the metadata that we write)

jsimonclark:
The pre-final fixes agreed in #105/#106 plus the loose ends from the #96 review and the reference-manifest audit. 

## What's in it

1. **`battinfo_dataset` record (#105).** BattINFO's dataset schema joins the bundle (fetched at the already-pinned upstream ref `b9b49bb`; verified self-contained, so nothing else joins). `Metadata` gains the record, exported as `bdf.BattinfoDataset`, with a `METADATA` target root. Nothing fills it automatically yet: it's the documented landing spot for file-level provenance (#92). Like every record it mirrors a pre-1.0 upstream schema and may evolve with it, as stated in the changelog.
2. **Writer version stamps (#106).** Every sidecar `save()` writes carries `bdf.bdf_version`, `bdf.ontology_version`, `bdf.battinfo_ref`. Stamps identify the writer: they overwrite on each save, never mutate the caller's object, and never conjure a sidecar for an empty `Metadata`.
3. **Version-aware sidecar errors.** A reserved sidecar that fails validation raises `BDFMetadataError` (one exception for every unusable-sidecar case); when the document claims a newer `bdf_version`, the message says which version wrote it and to upgrade. Malformed claimed versions fall through safely. The plugin-facing `JsonSidecarParser` is unchanged.
4. **Windows test fixes.** `re.escape` on the refuse-on-sidecar path match; explicit-LF fixtures for the two whole-head raw-capture tests.
5. **Reference-manifest integrity.** Ten of thirteen `datasets.json` sha256s hashed a Windows CRLF rendering and failed to verify on Linux. Now: shas hash stored bytes, `.gitattributes` pins the artifacts `-text`, and a gate test enforces manifest/file agreement in CI. Plus `manifest_version`, an `artifact_record` DOI placeholder for the canonical Zenodo record, and a corrected Stanford unconverted-file reason.

## Heads-up for existing checkouts

After pulling commit 5, a pre-existing clone shows all 13 reference artifacts as modified and the new sha test fails locally: stale line endings, not corruption. `git checkout -- docs/examples/reference` fixes it (noted in CONTRIBUTING).

## Sequencing

rc2 installs cannot read sidecars written after this lands (`extra="forbid"` meets the new stamp keys), which is exactly the failure mode item 3 exists to explain — recommend a short rc3 before final. The `artifact_record.doi` back-fill lands as one extra commit once the Zenodo draft's DOI is reserved.

Closes #106. Implements the #105 decision (dataset in, spec halves out);


pghege (approved):
Verified end to end: 879 unit tests pass locally, sidecar stamps behave exactly as documented (overwrite per save, never mutate the caller's object, no sidecar conjured for an empty Metadata, stamps carry bdf/ontology/battinfo versions), the dataset record is self-contained at the pinned ref, and the published Zenodo record (10.5281/zenodo.22230066) matches the PR's claims: 11 clean artifacts + manifest + checksums, no __Bug fixtures, DOI recorded in datasets.json. The anomaly note in the record description reads as the resolution-limited files from #98, consistent with the clean-set rule. Nice closing move for the record/repo linkage.

